Issue #9, opened 2015-04-19
Mismatch between wire format and logical data


vojtechhabarta:
This applies at least to enums and dates, enums are transferred as string, dates are transferred as string or number but it might be useful to work with original enum and date types.
Which TypeScript declarations should be generated? For data sent over the wire or for data we want to work with?

vojtechhabarta:
Dates can be mapped as `Date`, `number` or `string` based on user preference.
Enums are mapped as union of string literal types (`type Direction = "North" | "East" | "South" | "West";`)

PR #10, opened 2015-08-13
add support form enums and Java 8 Date API

zouabimourad:
Hello ! 

Thank you for your project.

I have done some modifications.

- Enums support  ( optional ) 
- Java 8 Date API support ( map LocalDateTime , LocalTime ... to Date ) 

If you are interrested .. merge !

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,8 +63,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.2</version>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
                     <showWarnings>true</showWarnings>
                     <showDeprecation>true</showDeprecation>
                     <compilerArgument>-Xlint</compilerArgument>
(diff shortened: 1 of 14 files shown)


vojtechhabarta:
Hello @zouabimourad,

sorry for late response. Thank you for the pull request, it's good to see interest in this project.

I have some comments on this PR.

Note regarding terminology: I think "ambient" word is used also for internal modules (http://www.typescriptlang.org/Handbook) and what you wanted was external module. So instead of "ambientModuleName" I would use something mentioning "external" like "useExternalModule", "emitExternalModule" or just "externalModule". 

Regarding support for Java 8 Date API: Are you able to implement this without requiring Java 8 using reflections and Java 7? I really want to support Java 7 projects for some time.

Regarding enums: Problem with enums here is that enum values are send as strings in JSON but TypeScript enum values are numbers. That's why these fields are declared as strings. I mentioned this also in issue #9. But I agree that it could be useful to declare these fields as enums and ensure that JSON strings received from Java server are transformed to enums and vice versa.

Regarding code formatting: It is not written anywhere but please use 4 spaces as indentation in this project and avoid reformatting existing code otherwise diff is huge and not readable.

This PR combines more features together, maybe it would be better to create separate PRs for individual features so we can solve it one by one.

vojtechhabarta:
@zouabimourad I have some comments on typescript-generator features.
- Now it is possible to implement custom `TypeProcessor` to handle Java 8 date/time classes.
- Namespace/module customization was improved.
- For better enum types mapping I wait for [String literal types](https://github.com/Microsoft/TypeScript/pull/5185) feature. This feature should be included in TypeScript 1.8 ([Roadmap](https://github.com/Microsoft/TypeScript/wiki/Roadmap))

PR #13, opened 2015-11-25
Ryan10132/add features

ryan10132:
Hi!

The goal of my branch is to be purely additive, if I didn't succeed in doing that, and I broke a use case you had just say so and I'll try to remedy that. I added the following.

Custom type parsing
  A user can override the way that certain types get parsed into TsTypes
Enums
  I was slightly opinionated about how I implemented this, but one way that I've found that works well
  is to create a variable with string constants on it
Generics
  Custom types that can have generics (e.g. MyCustomType<String, Integer>) are now output correctly.
  extends, super, ? and a few other java oddities are not intended to be supported, I found they did semi 
  reasonable things though
OutputStream as destination
  User can now give an output stream, which is more general than a file
Minor customization options
  Default indentation level, TsType name prefix, etc

Let me know if you have any questions at all! Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ release.properties
 /sample-gradle/.gradle/
 /sample-gradle/.nb-gradle/
 /sample-gradle/build/
+**.classpath
+**.project
+**.settings
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ release.properties
 **.classpath
 **.project
 **.settings
+*.d.ts
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/.gitignore
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/.gitignore
@@ -1,0 +1,1 @@
+example/
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/JavaToTypescriptTypeConverter.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/JavaToTypescriptTypeConverter.java
@@ -1,0 +1,10 @@
+package cz.habarta.typescript.generator;
+
+import java.lang.reflect.Type;
+
+public abstract class JavaToTypescriptTypeConverter {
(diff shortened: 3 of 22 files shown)


ryan10132:
I'm not sure why this is failing, the relevant line is probably 
ERROR] Failed to execute goal com.github.github:site-maven-plugin:0.11:site (site) on project typescript-generator: Server 'github' not found in settings -> [Help 1]

But I don't know what settings object is being referred to

vojtechhabarta:
@ryan10132 AppVeyor build is not probably correctly set up to build pull requests, don't worry about that.
Your PR is pretty big I will review it later.

vojtechhabarta:
@ryan10132 I like most of the features in this PR. However we need to solve some issues first.

One thing which I don't want to add right now is **Enum** support. I think it is better to wait for [String literal types](https://github.com/Microsoft/TypeScript/pull/5185) feature and see how this can help to solve "Enum" problem in a clean way. This feature should be included in TypeScript 1.8 ([Roadmap](https://github.com/Microsoft/TypeScript/wiki/Roadmap)).

Currently this PR is **not compilable**. You can see log [here](https://ci.appveyor.com/project/vojtechhabarta/typescript-generator/build/1.2.111-gelhe47a) (should be publicly visible). Compilation problem is somewhere in the middle. Compiler could not find symbols `Lists` and `Joiner`. This project currently doesn't have any dependencies except for JSON libraries so try to avoid adding new dependencies. The build didn't stop on maven error and instead failed on something else. I know this is not ideal, sorry for that.

Your PR **is large** and addresses many different features. Could it be possible to submit PRs one by one for individual features like _Custom type parsing_, _Generics_, _Minor customization options_?

ryan10132:
Thanks for taking a look! I should be able to split it up. 
I'm on board with typescript enums with strings, but until then can we have my current scheme hidden behind a flag? Something I would add to Settings that was disabled by default. When typescript added support, it could be removed.

vojtechhabarta:
This PR is superseded by subsequent smaller PRs.

PR #14, opened 2015-11-30
Ryan10132/add api enhancements

ryan10132:
Adds optional ts types
Adds custom ts type parsing
Adds output stream support for output

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@ release.properties
 /sample-gradle/.gradle/
 /sample-gradle/.nb-gradle/
 /sample-gradle/build/
+**.classpath
+**.project
+**.settings
+*.d.ts
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/.gitignore
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/.gitignore
@@ -1,0 +1,1 @@
+example/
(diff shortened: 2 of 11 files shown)


ryan10132:
Hi, I'm not sure why this is failing

vojtechhabarta:
@ryan10132 part of the build are "upload maven site" and "tag build" steps. These steps do not work for PR builds. I'm trying to improve the build. Your code is compilable see line [2042](https://ci.appveyor.com/project/vojtechhabarta/typescript-generator/build/1.2.114-xub0vm6i#L2042) in build log.

Thanks for splitting the PR. Could you please just remove unnecessary changes in import statements? For example `GenerateTask.java` contains no real change and there are 20 lines changed. This is most obvious case but these extra changes are in many java files. In this project import statements are sorted alphabetically and wildcard imports can be used.

ryan10132:
Alright, got it. I removed the unnecessary import changes. 

I'll run mvn test locally and decide if things are passing based on that in the future, thanks!

vojtechhabarta:
Thanks for removing diffs in imports.
Meanwhile I improved build so it should work for PRs.

In current codebase it's not easy to implement and test custom Java to TypeScript type mapping. I think root cause is `ModelCompiler.typeFromJava()` method which is hard to reuse and its contract is confusing. It serves two purposes: 1. maps Java type to TypeScript type and 2. discovers classes used in Java type which should be also processed. This is not obvious from method signature which uses _output_ parametr.

So I will merge your PR and then try to refactor this method together with your changes into more flexible and clean code.

ryan10132:
Cool, thanks, let me know what you come up with. I'll have another PR within a few hours for my next feature.

PR #15, opened 2015-12-02
Ryan10132/add style configuration

ryan10132:
Added several configuration options

Output is now sorted (this removes the possibility of non deterministic output, which is the current state)

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ModelCompiler.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ModelCompiler.java
@@ -103,12 +103,12 @@ public class ModelCompiler {
         if (cls == null) {
             return null;
         }
-        final String name = cls.getSimpleName();
+        String name = cls.getSimpleName();
         if (settings.removeTypeNameSuffix != null && name.endsWith(settings.removeTypeNameSuffix)) {
-            return name.substring(0, name.length() - settings.removeTypeNameSuffix.length());
-        } else {
-            return name;
+            name = name.substring(0, name.length() - settings.removeTypeNameSuffix.length());
         }
+        name = settings.addTypeNamePrefix + name;
+        return name;
     }
 
     private TsType replaceTypes(TsType type, LinkedHashSet<TsType.EnumType> replacedEnums, LinkedHashSet<TsType.AliasType> typeAliases) {
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -15,4 +15,5 @@ public class Settings {
     public String addTypeNamePrefix = "";
     public String addDeclarationPrefix = "";
     public int initialIndentationLevel = 0;
+    public boolean sortDeclarations = false;
 }
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -12,4 +12,7 @@ public class Settings {
     public String removeTypeNameSuffix = null;
     public DateMapping mapDate = DateMapping.asDate;
     public TypeProcessor customTypeProcessor = null;
+    public String addTypeNamePrefix = "";
+    public String addDeclarationPrefix = "";
+    public int initialIndentationLevel = 0;
 }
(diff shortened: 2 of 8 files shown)


vojtechhabarta:
Sorting output declarations: usually properties in classes have some logical order and I think it is useful to retain original order. But I understand that sometimes it is more useful to sort them. So could you please add `sortDeclarations` setting with `false` default value for this feature?

What is the purpose of `Settings.declarationPrefix`?

I would like to rename `defaultCustomTypePrefix` to `addTypeNamePrefix` which would be more consistent with `removeTypeNameSuffix` setting.

ryan10132:
I agree that sometimes properties have a logical order, however I was seeing nondeterministic behavior with the output of classes (I believe `BeanHelper.getProperties` was to blame). I'll make it optional anyway though in case this gets fixed.

`Settings.declarationPrefix` prefixes everything declared in a module. One way I was using this was to add an `export` to prefixes (I was putting the autogenerated types in a .ts file). So 
    interface A { }
might become
    export interface A {}
with `Settings.declarationPrefix = "export ";`
I think it should be renamed to `addDeclarationPrefix` for consistency, I'll do that

I will go ahead with the `defaultCustomTypePrefix` rename that you suggested

ryan10132:
oops accidentally closed

vojtechhabarta:
**Sorting**
I tried some variants and I got properties in original order. But I think it is really good feature anyway.

**Declaration prefix**
Purpose of this tool is to generate *.d.ts file which I think is proper solution to describe JSON data.
I know it might be useful to generate *.ts file but this is more complicated. We can discuss it in the future.
If we decide to support this scenario then tool should be able to generate correct *.ts files itself without any user input.
What do you think?

ryan10132:
I see, I still think declaration prefixes have merit in a *.d.ts file (for
example, you might want to prepend "declare", or you might want to prepend
a comment (perhaps you want a warning on every declaration), something like
```
// This was autogenerated, don't modify it
interface A { }
```
In that case, `Settings.addTypeNamePrefix = "// This was autogenerated\n";`


One of the reasons that I added support for using an OutputStream in the
previous CR was so that I could have more flexibility about where my type
definitions were written to. My needs are for outputting .ts files
something like this

```
module foo {

// type declarations from typescript-generator

// service interface from another generator

// service class from another generator

}
```
And in that case I am only using your library to generate the first comment
block, not the entire file.

On Thu, Dec 3, 2015 at 1:49 AM, vojtechhabarta <notifications@github.com>
wrote:

> *Sorting*
> I tried some variants and I got properties in original order. But I think
> it is really good feature anyway.
>
> *Declaration prefix*
> Purpose of this tool is to generate *.d.ts file which I think is proper
> solution to describe JSON data.
> I know it might be useful to generate *.ts file but this is more
> complicated. We can discuss it in the future.
> If we decide to support this scenario then tool should be able to generate
> correct *.ts files itself without any user input.
> What do you think?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vojtechhabarta/typescript-generator/pull/15#issuecomment-161570657>
> .
>



-- 
Ryan McNamara
Senior at Michigan Technological University
Seeking Bachelors in Computer Science and Mathematics
rjmcnama@mtu.edu

ryan10132:
Tldr is that I think having the declaration prefix is ok because it doesn't give the library any indication of what kind of file it is generating, and I think its contract is simple enough

vojtechhabarta:
Thanks for your explanation. Now I understand your use case.
I still don't want to allow user to add arbitrary texts into output. But I would like to have direct support for your use case and I think I have good solution for it:

1. Add `export` keyword to appropriate places so that declarations are usable in both *.d.ts and *.ts files.
2. Add some boolean option like `emitEmbeddableOutput` to settings. This would tell typescript-generator that it is running as part of another generator (which emits namespace/module instead of it) and that it should just emit type declarations with right indentation. No need to configure texts.

Does it sound reasonable for you? If yes, I would implement it.

vojtechhabarta:
BTW from TypeScript 1.5 they recommend to use `namespace X {` instead of `module X {` (https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Namespaces%20and%20Modules.md)

vojtechhabarta:
I was already thinking about adding some comment maybe even with timestamp and used typescript-generator version.

ryan10132:
Yea that sounds pretty good. 
For 1, are you suggesting that we add export to every appropriate declaration? That sounds fine to me, should it be behind a flag?

For 2, I don't actually think this is necessary, I can achieve this already by changing the indentation in the settings and not specifying a module / namespace, with these two settings it does exactly what you say -> only emits the type declarations, indented appropriately, and without the namespace / module

vojtechhabarta:
Yes, specifying module/namespace and them omitting its declaration is weird.

I will add special method in `TypeScriptGenerator` for this use case so these options are not part of the main API.

vojtechhabarta:
I added `TypeScriptGenerator.generateEmbeddableTypeScript()` method instead of `Settings.addDeclarationPrefix` and `Settings.initialIndentationLevel`. Tests passed but please check if everything is ok.

ryan10132:
First glance looks good! thanks

PR #16, opened 2015-12-04
Added means of getting a context for better type conversion support

ryan10132:
I (the client) don't want to have to create a `TypeProcessor.Context` to convert a type. Specifically, my reasoning is that there's some potentially complicated logic in implementing getMappedName (probably going to get even more complicated when I add generics)


vojtechhabarta:
After TypeScriptGenerator refactoring is it possible to use `new TypeScriptGenerator(settings).getModelCompiler().typeFromJava()`. I think this is what you need, please comment if you need additional support.

PR #17, opened 2015-12-05
Ryan10132/add generics

ryan10132:
Add support for generics. Extends and super are not expected to work.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/DefaultTypeProcessor.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/DefaultTypeProcessor.java
@@ -4,9 +4,11 @@ package cz.habarta.typescript.generator;
 import java.lang.reflect.*;
 import java.util.*;
 
+import cz.habarta.typescript.generator.TsType.*;
+
 
 public class DefaultTypeProcessor implements TypeProcessor {
-    
+
     @Override
     public Result processType(Type javaType, Context context) {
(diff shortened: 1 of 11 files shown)


vojtechhabarta:
@ryan10132  I took your PR as an inspiration and implemented some changes. By default typescript-generator now supports generic classes by removing type arguments (Java class `A<T>` is mapped to TypeScript interface `A`).

But I also implemented your generics feature as an example of custom `TypeProcessor` in `GenericsTest`. I wanted to test TypeProcessor concept if it is flexible enough and indeed it required some changes in core to make it work. You can take this custom TypeProcessor from test and use it in your code base.

ryan10132:
Can we expose the real generic handling code in some way? Basically anyone who wants real generic support is going to have to write it themselves (like you did in GenericsTest). Even just exposing a custom type processor that handles generics appropriately (like the one created in GenericsTest) should be enough.

vojtechhabarta:
I would appreciate some use case for generics usage in JSON data other than in `Collection` and `Map` classes.

ryan10132:
My use case is something like this: 

I have an endpoint that returns a paginated response. That json response looks like 

```
interface Page<T> {
    payload: T[];
    pageId: number;
    nextPageId: number;
}
```

I could imagine a use case for a graph representation:

```
interface Graph<E, V> {
    edges: Edge<E>;
    vertices: V[];
}
```

Another example would be a linked list that was actually serialized like one (not just a Collection linked list), though I suppose this is special case of a graph

vojtechhabarta:
Yes, Page<T> looks like good reason. I created public class `GenericsTypeProcessor`.

PR #18, opened 2015-12-07
Several bug fixes

ryan10132:
Added void.class and Void.class as known types.
Enum conversion now works (calling to string on an enumType now
correctly returns "string")
Made ModelCompiler.join public

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/DefaultTypeProcessor.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/DefaultTypeProcessor.java
@@ -74,6 +74,8 @@ public class DefaultTypeProcessor implements TypeProcessor {
         knownTypes.put(Character.TYPE, TsType.String);
         knownTypes.put(String.class, TsType.String);
         knownTypes.put(Date.class, TsType.Date);
+        knownTypes.put(void.class, TsType.Void);
+        knownTypes.put(Void.class, TsType.Void);
         return knownTypes;
     }
 
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
@@ -11,6 +11,7 @@ public abstract class TsType {
     public static final TsType Number = new BasicType("number");
     public static final TsType String = new BasicType("string");
     public static final TsType Date = new BasicType("Date");
+    public static final TsType Void = new BasicType("void");
 
     public static final AliasType DateAsNumber = new AliasType("DateAsNumber", "type DateAsNumber = number;");
     public static final AliasType DateAsString = new AliasType("DateAsString", "type DateAsString = string;");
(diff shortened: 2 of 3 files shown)


vojtechhabarta:
`ModelCompiler.join` should not be public. It is tiny little utility method which should not be part of typescript-generator API. Clients of this project can use `Collectors.joining()` in Java 8 or `Joiner` from guava.

Couple of questions:
What is the reason for `void` type?
What is wrong with enum conversion?

ryan10132:
I totally agree that ModelCompiler.join shouldn't be public.
 
Didn't know about Collectors.join, thanks!

The void type allows me to convert `Void.class` or `void.class` to typescripts string representation for void. It *is* a valid type in java, and it is a primitive type so it should be handled specially just like `int` or `long`. Specifically I run in to this when trying to convert the return type of a method, which is sometimes void.

Enum conversion had an issue where if you called `compiler.convertType(enumType)` you would get the enum type back, but calling `toString` would produce the original type and *not* `"string"` as desired. Since it seemed that the `toString`s of all the other `TsType`s returned the desired representation, this seemed more consistent.

I added a couple tests to make this more obvious

vojtechhabarta:
Yes, `void` is valid Java type. (BTW it is not considered [primitive data type](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html)) Sure, method can return `void` but then it is not property getter and doesn't represent JSON data. I am just curious for what you want to use typescript-generator.

For `EnumType`: You are right that currently `"string"` would be more consistent. But `EnumType` has special handling in `Emitter` and it works as currently expected. The reason I don't want to change it now to `string` is that from TypeScript 1.8 we would probably want to change it back to name of the enum. Enum type declaration will probably look like this:
```
type Direction = "North" | "East" | "South" | "West";
```
and then usage will be as follows:
```
interface A {
    direction: Direction
}
```
So we will need `EnumType.toString()` to return `"Direction"` in this case.

ryan10132:
Part of my use case is that I want to convert a bunch of types using `ModelCompiler.typeFromJava`, outside of just generating. The reason I want this is that I want to be able to output typescript method declarations (I wrote a tool on top of yours to do this), so that's why I would like void to convert nicely, it isn't part of a JSON definition or something weird like that.

I definitely agree that *when* typescript has string based enums, it should output those, but until then I think it's worth being consistent.

vojtechhabarta:
Actually I was not exactly correct. `EnumType` is replaced by `TsType.String` in `ModelCompiler`. There is no need for special handling in `Emitter` and I removed it (it was left there accidentally).
BTW this replacement is recursive and your change breaks "Original type" comment when enum is used inside of some container type (array or map).

So `EnumType.toString()` should not be changed but instead you will probably suggest to make `ModelCompiler.replaceTypes()` functionality available. Am I right?

ryan10132:
I see. I'm not sure what the right solution is. Perhaps the right thing for me to do for now (until typescript has real enums) is hack this together on my side. I think I'll also output enums in my own custom format for now as well and not try to push this repo to adopt my strategy temporarily.

vojtechhabarta:
I added `ModelCompiler.typeFromJavaWithReplacement()` so now there are two methods: one returns original type (for `SomeEnum` it returns `"SomeEnum"`) and second returns type after replacement (for `SomeEnum` it returns `"string"`). Hope this helps. (There is also test.)

Issue #20, opened 2015-12-10
Automatic class discovery from REST resources


mzuzcak:
please add automatic discovery of rest classes, based on consumed and produced classes from resources.

vojtechhabarta:
Great idea.
For example from this REST resource class:
```
@Path("user")
public class UserResource {
    @GET
    public User getCurrentUser() {}
}
```
we can discover JSON class `User`.

But we need to first discover resource classes. This could be done using some classpath scanning library or maybe we can get list of resources from JAX-RS/Jersey.

Classpath scanning is slow but this feature is meant to be turned on by user so I think this is not a problem.

vojtechhabarta:
Implemented automatic discovery of JSON classes used in configured JAX-RS application (`classesFromJaxrsApplication` option). It is possible to exclude some resources or classes from processing using `excludeClasses` option. Also practical is to set `sortTypeDeclarations` option to sort generated interfaces alphabetically.

(part of 1.4.152 release)

PR #21, opened 2015-12-14
ability to specify which files to convert also with a glob pattern instead of listing all the classes

emmanueltouzery:
this maven plugin looks great, however for bigger applications, it will quickly become a burden to list all the DTOs. With this change it's possible to simply specify a glob pattern for the classes to convert and the plugin finds them itself.

I have concerns about the implementation, it's the first time I touch a maven mojo... To find the classes, I walk the directory tree, and then have a little function to guess the class name from the file name. I would expect I could enumerate the classes from the classloader, but it seems to be possible [only through hacks](http://stackoverflow.com/questions/2681459/how-can-i-list-all-classes-loaded-in-a-specific-class-loader). I also thought about using the `reflections` library, but it seems overkill to add that new dependency to your mojo only for that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ In Maven build you can use `typescript-generator-maven-plugin` like this:
 </plugin>
 ```
 
+Instead of listing all the classes, you can also define a list of glob patterns patterns matching the classes to export, like that for instance:
+``` xml
+<classPatterns>
+    <classPattern>*TsDto</classPattern>
+</classPatterns>
+```
+
 More complete sample can be found [here](sample-maven).
 Detailed description how to configure typescript-generator-maven-plugin is on generated [site](http://vojtechhabarta.github.io/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html).
 
@@ -95,7 +102,7 @@ generateTypeScript {
 ```
 
 More complete sample can be found [here](sample-gradle).
-Gradle plugin has the same features as Maven plugin, for detailed description see Maven generated [site](http://vojtechhabarta.github.io/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html). 
+Gradle plugin has the same features as Maven plugin, for detailed description see Maven generated [site](http://vojtechhabarta.github.io/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html).
 
 
 Direct invocation
@@ -113,7 +120,7 @@ Architecture
 ```
            (Model)            (TsModel)
 ModelParser  ==>  ModelCompiler  ==>  Emitter
-         |         | 
+         |         |
          V         V
         TypeProcessor
 ```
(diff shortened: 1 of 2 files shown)


emmanueltouzery:
I just spotted https://github.com/vojtechhabarta/typescript-generator/issues/20 -- I think that in any case rather than discovery by tying the plugin to a framework (eg JAX-RS) or searching for some annotations, some glob pattern approach (eg *TsDto or *Dto classes) is going to be much faster and also more powerful -- eg maybe the user doesn't want to convert all the DTOs.
For instance, many applications will use REST services to communicate with the browser (so typescript client) but also some of the REST URLs would be to communicate with other apps, using REST for interoperability, and in that case the typescript interfaces would be generated without being needed.

vojtechhabarta:
It is not needed to list all JSON classes, it is sufficient to list "root" classes and typescript-generator will process all classes used in these "root" classes. But you are right for bigger applications number of "root" classes grows. That's the reason why I started to implement class discovery from JAX-RS application (#20). JAX-RS is specification with few implementations and is used in many cases. But I agree it is not universal way how to get JSON classes so typescript-generator could definitely support class patterns.

However I think implementing file name patterns in maven plugin is not right way to do it for these reasons:
1. It is not possible to find classes in dependencies. For example when one module contains JSON DTOs and second contains REST and typescript-generator.
2. Mapping file name to class name is fragile. This could be acceptable but we could do it better.
3. I am also not comfortable with writing Maven plugins and even less with writing Gradle plugins. So these build plugins should remain thin layer on top of `typescript-generator-core`. Also such a good feature should not be available only in Maven but also in Gradle and programmatically.

My proposal is to implement class name globbing using some classpath scanning library (maybe guava [ClassPath](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/reflect/ClassPath.html)). I don't like adding dependencies but in this case it is probably better than re-implementing classpath scanning in typescript-generator. This globbing feature should be implemented in `core` and used from both build plugins. Since #20 contains refactoring of input classes this should be implemented after #20. What do you think?

emmanueltouzery:
I agree with all points, most of all I don't know how could I miss point 1.
One thing I am not convinced of is to start from JAX-RS root classes instead of directly finding the DTOs, for performance and code complexity reasons, including what I mentionned that I think that would in any case return too many classes and then you have to glob them, in some cases. But I have little experience with classpath scanning snd obviously it's used a lot and performance is not that big of an issue.

You said you started implementing #20. Any idea when could you have it finished? Thank you!

emmanueltouzery:
Note that in our project we use JAX-RS so we are covered by this plan, and it's good enough for us, no problem. However, in general, I believe apache CXF is an example of a competing API that would not be covered.

vojtechhabarta:
I hope I will finish #20 this week.

I agree that class patterns are useful and that not all use cases involve JAX-RS application. But we can have both features. (BTW Apache CXF also implements JAX-RS)

emmanueltouzery:
ok thanks for the info. Closing the PR then, glad to read about the upcoming fix for #20!

PR #25, opened 2016-01-22
New test illustrating 'shortcoming' of the generator

ryan10132:
I'd like for the typescript-generator generic processor to support wildcards.
The test I added fails on purpose to show what I'd like the behavior to be, currently it NPEs

My specific use case is an object where I have a List<? extends Exception> that I want serialized.

Let me know what you think! Thanks.


vojtechhabarta:
Thanks for this bug report / feature request. typescript-generator definitely should not crash on any input.

I implemented wildcard types processing as specified in the test.

PR #27, opened 2016-01-22
Added equals for tsType and made toString mandatory

ryan10132:
My use case here is to make it easier to test the library that depends on this one.

By the way, I created a typescript service generating library that is based of yours! Check it out and let me know what you think!

https://github.com/palantir/typescript-service-generator/

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
@@ -16,9 +16,22 @@ public abstract class TsType {
     public static final AliasType DateAsNumber = new AliasType("DateAsNumber", "type DateAsNumber = number;");
     public static final AliasType DateAsString = new AliasType("DateAsString", "type DateAsString = string;");
 
+    @Override
+    public boolean equals(Object rhs) {
+        return rhs != null && this.getClass() == rhs.getClass() && this.toString().equals(rhs.toString());
+    }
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
I didn't have a chance to look at typescript-service-generator yet but in general I am happy that typescript-generator is used directly or in other tools.

vojtechhabarta:
Please add these tests and make them work:
``` java
assertEquals(new TsType.StructuralType("Foo"), new TsType.StructuralType(new String("Foo")));
assertFalse(new TsType.StructuralType("Foo").equals(null));
```

Also please add `hashCode` method consistent with `equals`.

ryan10132:
sure thing, done

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
```diff
@@ -16,10 +16,23 @@
     public static final AliasType DateAsNumber = new AliasType("DateAsNumber", "type DateAsNumber = number;");
     public static final AliasType DateAsString = new AliasType("DateAsString", "type DateAsString = string;");
 
+    @Override
+    public boolean equals(Object rhs) {
+        return rhs != null && this.getClass() == rhs.getClass() && this.toString() == rhs.toString();
```
vojtechhabarta: You should not use `==` for string comparisons. That was the first test line I suggested to add.

PR #30, opened 2016-03-09
Allow annotations to determine optionality

tjwilson90:
This allows methods and fields that are annotated as @Nullable (or any
other user chosen annotation) to generate optional fields in your
typescript definitions.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -3,6 +3,7 @@ package cz.habarta.typescript.generator;
 
 import cz.habarta.typescript.generator.emitter.EmitterExtension;
 import java.io.File;
+import java.lang.annotation.Annotation;
 import java.util.*;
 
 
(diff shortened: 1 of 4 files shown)


vojtechhabarta:
Thanks for this PR, I had some use case for optionals just this week :smile:

One question: What is the purpose of `Jackson[1|2]Parser(Settings, TypeProcessor, ObjectMapper)` constructors? They are not used in typescript-generator so I guess you are using them elsewhere or plan to do so. Issue with this is that typescript-generator API entry point is `TypeScriptGenerator` class and dependent types. Other classes can be refactored and cleaned in the future. So I can improve `TypeScriptGenerator` class somehow to support your use case.

Second thing is that in general all features should be accessible also using maven and gradle plugins. I will do this after merge.

tjwilson90:
I could see customizing the object mapper used by the jackson parser as a useful feature. For example, setting up the AnnotationIntrospector for your Jackson1Parser so it would consider all annotations.

I don't actually plan on using the Jackson1Parser, so I don't need this to be supported. Feel free to implement this part differently.

PR #32, opened 2016-03-18
Changes emitEnums to emit 'enum' instead of 'type'; adds export keyword to modules, and declare keyword to ambient files.

lonnelars:
`typescript-generator` outputs type aliases for java enums, which I believe is wrong, according to the [spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.10). 

This PR proposes to solve the problem by changing `type` to `enum`. It also adds the `export` keyword for `outputFileType=module`, and the `declare` keyword for other output file types. 

This works for our setup, which produces `.ts` files from Java classes and enums. However, I might have missed other use cases.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/emitter/Emitter.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/emitter/Emitter.java
@@ -123,17 +123,24 @@ public class Emitter {
 
     private void emitEnums(TsModel model, boolean exportKeyword) {
         final ArrayList<TsEnumModel> enums = new ArrayList<>(model.getEnums());
+        if (!enums.isEmpty() && settings.outputKind == TypeScriptOutputKind.module && settings.outputFileType == TypeScriptFileType.declarationFile) {
+            throw new IllegalArgumentException("Tried to output an enum with a top-level export declaration in a " +
+                    "declaration file. This is not valid TypeScript. Either change the outputFileType to " +
+                    "implementationFile, or use another outputKind.");
+        }
         if (settings.sortDeclarations || settings.sortTypeDeclarations) {
             Collections.sort(enums);
         }
         for (TsEnumModel tsEnum : enums) {
             writeNewLine();
-            final ArrayList<String> quotedValues = new ArrayList<>();
-            for (String value : tsEnum.getValues()) {
-                quotedValues.add(settings.quotes + value + settings.quotes);
-            }
             emitComments(tsEnum.getComments());
-            writeIndentedLine(exportKeyword, "type " + tsEnum.getName() + " = " + ModelCompiler.join(quotedValues, " | ") + ";");
+            String exportOrDeclare;
+            if (exportKeyword) {
+                exportOrDeclare = "export";
+            } else {
+                exportOrDeclare = "declare";
+            }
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
@lonnelars thanks for this PR, I have some comments:

Enums
---
Problem is that Java enum values are **strings** in JSON but TypeScript enums are **number** based.

When we have this Java code
``` java
class AClass {
    public Direction direction;
}
enum Direction { North, East, South, West }
```
produced JSON is in this format
``` json
{
    "direction": "North"
}
```
so in TypeScript generated type of "direction" field should be `string` or `"North" | "East" | "South" | "West"` (this feature is available in TypeScript from version 1.8).
I can imagine that someone first replaces string values with enum values in JSON data and then use modified data but this is something else.

Modules
---
Actually I am not sure what you wanted to change. Can you look at the table on [Modules and Namespaces](http://vojtechhabarta.github.io/typescript-generator/doc/ModulesAndNamespaces.html) page if there is or isn't variant you want add or change? There is a variant with `outputFileType = implementationFile`, `outputKind = module` which you might want to check.

The table is complicated but it shows different variants of module / namespace combinations in declaration (.d.ts) and implementation (.ts) files that are supported by typescript-generator. There is also test in class `cz.habarta.typescript.generator.ModulesAndNamespacesTest` which produces examples of all these 10 variants. I just added enums to these examples. You can check produced files in `typescript-generator-core/target/test-ts-modules` directory. And you can also switch to `typescript-generator-core/src/test/ts` directory, run `tsc` compiler and check if it runs without errors.

lonnelars:
First of all, thanks for fixing top-level exports in the latest version, we really appreciated it :+1: 

The problem we have now is the following:
1. We have an `enum` in the java API that we need to export to typescript. 
2. Using v1.6.171, the `enum` is exported as `export type MyEnum = "value1" | "value2" ...`. This does not compile, and gives me the error `TS1110: Type expected.`
3. What I tried to solve with the PR, was to change the export of the java enum to a Typescript enum. This requires that the output file is an implementation file, and that's why I included the check
```java
if (!enums.isEmpty() &&
    settings.outputKind == TypeScriptOutputKind.module && 
    settings.outputFileType == TypeScriptFileType.declarationFile)
```

This PR works for our needs, but I realize that it probably is not general enough for all use cases.

vojtechhabarta:
@lonnelars error you mentioned is most probably produced by TypeScript version earlier than 1.8.

For `enum Direction { North, East, South, West }` in implementation file TypeScript compiler outputs
``` typescript
var Direction;
(function (Direction) {
    Direction[Direction["North"] = 0] = "North";
    Direction[Direction["East"] = 1] = "East";
    Direction[Direction["South"] = 2] = "South";
    Direction[Direction["West"] = 3] = "West";
})(Direction || (Direction = {}));
```
If you add `declare` keyword then mentioned code will not be generated and user will have problems using this enum. For example `var direction: string = Direction[Direction.East];` will compile but it will fail in runtime.

So in other words it is not useful to generate `declare enum Direction {North, East, South, West}` because user had to write `enum Direction {North, East, South, West}` anyway in some implementation file.

Current Java enum mapping to TypeScript describes data correctly and works in declaration files. But if you still want to generate TypeScript enums you can implement typescript-generator extension. See example in `cz.habarta.typescript.generator.ext` package.

lonnelars:
OK, I'll have a look at the generator-extension. Thank you for your feedback.

Issue #33, opened 2016-03-18
Better type safety with enums


ryan10132:
Loving the typescript 1.8 enums, gets me much better type safety while still having the wire format of the values be strings.

However I think we can do even better with type safety. Consider the following case, where the enum "Foo" was generated

```
type Foo = "a" | "b" | "c";

// ...

var x: Foo = "a"; // This is type safe
var xIsA = x === "a"; // This is not type safe
var xIsD = x === "d"; // This compiles, which is not ideal
```

Here is my idea:

```
type Foo = "a" | "b" | "c";
const FooValues {
    a: <Foo> "a",
    b: <Foo> "b",
    c: <Foo> "c"
}

// ...

var x: Foo = FooValues.a; // This is type safe
var xIsA = x === FooValues.a; // This is type safe
var xIsD = x === FooValues.d; // This does not compile, which is ideal
```

Writing such a "Values" const alongside every enum manually would be very tedious, but since it's all autogenerated, I think it's fine. Could also be behind a flag for people that don't want it.

What do you think? It would be very straightforward to implement, I'd be happy to submit a PR, would you accept it?

vojtechhabarta:
@ryan10132, having constants for enum values is nice however I have few comments.

#### Extension for implementation files
This only make sense in implementation files (.ts) and not in declaration files (.d.ts) since it generates JavaScript runtime code. So we can implement it as an extension, for example `cz.habarta.typescript.generator.ext.EnumConstantsExtension`.

#### Type safety
Assignment is type safe but comparison is not.
``` typescript
var xIsD = x === Foo.d; // This does not compile, which is ideal
// but
var xIsD = x === "d"; // This still compiles
var xIsD = x === Bar.d; // This also compiles (if Bar.d exists)
```

#### Names of const variable
We can even use the same name for `const` values variable.
``` typescript
type Foo = "a" | "b" | "c";
const Foo = {
    a: <Foo>"a",
    b: <Foo>"b",
    c: <Foo>"c"
}
```
First `Foo` is name of the type while second `Foo` is name of variable so they do not conflict because they are in different space.

#### Names of values
If value is for example `"a-a"` it cannot be used as value name, it is not a valid identifier. Currently this can't happen since value names are taken directly from Java enum names but for example Jackson has different options how to serialize enums so eventually it will happen.

#### Formatting
Try to be consistent with TypeScript formatter, there is no space in `<Foo>"a"`.

[...]

PR #34, opened 2016-03-19
Added enum constants extension

ryan10132:
Added extension that generates constants for enums.

### DIFF
--- a/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/EnumConstantsExtensionTest.java
+++ b/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/EnumConstantsExtensionTest.java
@@ -1,0 +1,58 @@
+
+package cz.habarta.typescript.generator;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import cz.habarta.typescript.generator.EnumTest.Direction;
+import cz.habarta.typescript.generator.emitter.EmitterExtension;
+import cz.habarta.typescript.generator.emitter.TsModel;
+import cz.habarta.typescript.generator.ext.EnumConstantsExtension;
+import cz.habarta.typescript.generator.parser.Jackson2Parser;
+import cz.habarta.typescript.generator.parser.Model;
+
+
+public class EnumConstantsExtensionTest {
+
+    @Test
+    public void basicTest() {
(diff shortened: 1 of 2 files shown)


ryan10132:
This resolves https://github.com/vojtechhabarta/typescript-generator/issues/33

vojtechhabarta:
@ryan10132 build of this PR failed. There is test compilation error caused by recent class rename. Can you fix it?

typescript-generator-core/src/test/java/cz/habarta/typescript/generator/EnumConstantsExtensionTest.java
```diff
@@ -0,0 +1,57 @@
...
+        Assert.assertEquals(indent + "West: <Direction>\"West\",", lines.get(4));
+        Assert.assertEquals("}", lines.get(5));
+        Assert.assertEquals("", lines.get(6));
+    }
+
+    @Test
+    public void testInTypeScriptGenerator() {
+        final Settings settings = new Settings();
+        settings.newline = "\n";
+        settings.outputFileType = TypeScriptFormat.implementationFile;
```
vojtechhabarta: This line is not compilable, should be `TypeScriptFileType.implementationFile`.

ryan10132: Sure, will fix later today

ryan10132:
Thanks! When should I expect to see it in a release? More generally, how do you decide when to cut a release / how often do you do it?

vojtechhabarta:
My current plan is to release new version each month (so next release in April).
Is that ok?

ryan10132:
I'm a bit blocked on this feature, any chance I could get a release now? I'm not horribly blocked so I could definitely live without it.

Also I'm curious why you want to do monthly releases and not more frequently. To reduce noise? More testing / vetting? 

My understanding is that github has a "click here to release" and then enter release notes and you're done, is that right?

vojtechhabarta:
I don't want to release too often because of several reasons. Some of them you mentioned. Next thing is that sometimes there is a combination of features which I want to release together. Improved stability when using it for some time before release. And of course it is some work: uploading artifacts to maven central repository, sometimes update documentation, release notes.

I will release it today (when you asked for it) but in general I don't plan to make frequent releases :smile:

ryan10132:
Got it, thanks for the exception, much appreciated! :)

PR #35, opened 2016-03-23
Highlighting generics bug

ryan10132:
Added an intentionally-failing test to highlight a bug. Classes of the form

```
class E extends D<string>  {
//...
```

Generate as

```
interface E extends D<T> {
//...
```

I'm not sure why, if I get some time I'll look into it later today


vojtechhabarta:
I was able to fix this bug, so I merged this test and committed the fix.

vojtechhabarta:
Released to Maven central repo ([v1.6.176](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.6.176))

PR #38, opened 2016-04-05
ModelCompiler refactoring

vojtechhabarta:
As typescript-generator is growing it is becoming harder to add new features. So I am refactoring ModelCompiler to have better structure.

`ModelCompiler` is now sequence of steps with clear contract, it:
- creates TypeScript model from Java model (`TsModel` -> `Model`) - this step uses Symbols instead of type names which are created in last step
- applies needed transformations like
    - dates to strings or numbers (`TsModel` -> `TsModel`)
    - enums to string literal unions (`TsModel` -> `TsModel`)
- resolves symbols to actual type names based on user settings

`Symbol`s represent TypeScript type when its name is not determined yet. It is indirection between Java type and TypeScript type.
`SymbolTable` is factory for symbols, has list of used symbols and it can report name conflicts in the future.
`TsBeanModel`, `TsEnumModel` and `TsAliasModel` now have common base class.

This refactoring is compatible from users perspective (Maven and Gradle plugins should work exactly as before) but the API is changing. If you are using it please review this change and report any missing functionality (mentioning some PR authors: @ryan10132 @lonnelars @tjwilson90).


tjwilson90:
I don't think I'll have any issues with this. Looks good.

ryan10132:
Looks good overall, I'll test it out by taking a dev dependency on it in my project

ryan10132:
Functionality that I seem to have lost is `ModelCompiler.typeFromJavaWithReplacement`

tjwilson90:
To convert a single type you can go from Type -> ModelParser.parseModel -> ModelCompiler.javaToTypeScript.

ryan10132:
That almost works, I think it will work for types that end up actually producing beans, or perhaps enums, however for built in types, at least, it isn't working:
image

In the above, javaType is `java.lang.String`

This doesn't surprise me, as I wouldn't expect the resulting tsModel to record that it found a String.

vojtechhabarta:
For tests I have `TestUtils.compileType()`. Could you try it? If it does what you need I will add some method to `ModelCompiler` with the same functionality.

ryan10132:
Yup, that works for me

PR #54, opened 2016-04-25
Support for ZonedDateTime.

thamacher:
I just included the ZonedDateTime to be handled like a util.Date.
I'd be glad, if you would add this change.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/DefaultTypeProcessor.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/DefaultTypeProcessor.java
@@ -3,6 +3,7 @@ package cz.habarta.typescript.generator;
 
 import java.lang.reflect.*;
 import java.math.*;
+import java.time.ZonedDateTime;
 import java.util.*;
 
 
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Typescript-generator is still on Java 7 so it is not possible to add support for Java 8 classes this way.
Currently it is possible to implement custom `TypeProcessor` which is little bit complicated mechanism for such simple customization so I plan to add possibility to specify custom type mappings in configuration.

thamacher:
Hi, 

I just had the same idea with the type-mappings. So my first
issue can easily be solved with that. I just added a new pull-request
with a first solution. 

Maybe you like it 

Thomas 

Am 25.04.2016
14:44, schrieb Vojtěch Habarta: 

> Typescript-generator is still on
Java 7 so it is not possible to add support for Java 8 classes this
way.
> Currently it is possible to implement custom TypeProcessor which
is little bit complicated mechanism for such simple customization so I
plan to add possibility to specify custom type mappings in
configuration. 
> 
> --
> You are receiving this because you authored
the thread.
> Reply to this email directly or view it on GitHub [1]



Links:
------
[1]
https://github.com/vojtechhabarta/typescript-generator/pull/54#issuecomment-214366806

PR #56, opened 2016-04-25
Support for customized TypeScript-Mapping

thamacher:
Hi Voji,

I just implemented some enhancements for your TypeScript-Generator, that enables me (and possibly other users) to define their custom-mappings.
We currently have a file with our common definitions for basic data (such as Money etc). As we already talked about the problem with using custom JsonSerializers and mapping them to TypeScript-Types, this would solve the problem for us now, as we can simply override the "any"-declaration with any type we would like to use.
Furthermore this also solves the problem with using JDK8 classes or joda-time, as this can simply be done by defining a custom-mapping.

E.g. adding the following properties to the config-file would already solve this:
`
references = [
    'references/references.ts'
  ]
  customMappings = [
    'org.joda.money.Money:Money',
    'java.time.ZonedDateTime:Date'
  ]
`
I hope you like the solution

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/CustomMappingTypeProcessor.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/CustomMappingTypeProcessor.java
@@ -1,0 +1,43 @@
+
+package cz.habarta.typescript.generator;
+
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class CustomMappingTypeProcessor implements TypeProcessor {
+
+    private final Map<String, String> customMappings = new HashMap<>();
+
+    public CustomMappingTypeProcessor(List<String> customMappings) {
+        if (customMappings != null) {
+            for (String customMapping : customMappings) {
+                this.customMappings.putAll(convertToMap(customMapping));
+            }
+        }
+    }
+
+    @Override
+    public Result processType(Type javaType, Context context) {
+        final Class<?> rawClass = Utils.getRawClassOrNull(javaType);
+        if (rawClass != null) {
+            String type = customMappings.get(rawClass.getName());
+            if (type != null) {
+                return new Result(new TsType.BasicType(type));
+            }
+        }
+
+        return null;
+    }
+
+    private Map<String, String> convertToMap(String mappingString) {
+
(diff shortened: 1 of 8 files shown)


vojtechhabarta:
Thanks for this PR, it is almost the same as what I was planning.

Originally I was thinking about "including" user types but "referencing" them is probably better. However this is not helpful when modules are used. So we should also provide possibility to add `import` statements. There are several syntaxes (see [MDN](https://developer.mozilla.org/en/docs/web/javascript/reference/statements/import) or [ECMAScript 2015](http://www.ecma-international.org/ecma-262/6.0/#sec-imports)). The question is if we should support one or two variants (similarly to your references) or to allow user to write whole import in any variant.

I wanted to specify mapping using `Map` class but I realized that
1. `Map` keys in Maven are represented using XML element names and XML element name cannot contain `$` character which is needed for nested classes.
2. `Properties` in Maven are too verbose.
So I think your format (`List<String>` with `:` character as delimiter) is better choice and I used it also for #57. I also created reusable method `Settings.convertToMap()`.

Currently this PR is not compilable because it contains commit with Java 8 classes (this is probably accidental because PR is created from your `master` branch, I think it is better to use different branches for PRs). Also please do not reformat Java imports (`GenerateTask` class), they should be sorted alphabetically in this project.

Do you want to fix this PR or may I copy&paste relevant parts and implement this feature?

thamacher:
Hi Voijta,

I am glad you liked my approach.
Please feel free to use and adjust my code.
Yes - I didn't want to merge these 2 requests. Actually I haven't committed to GitHub projects that much lately, so I'm not really used to it yet. I will try to do better next time.
Same with the change of imports - this was not intended, but done by my IDE automatically. I will take care better.

Concerning the import: I totally agree with you. This solution works for me right now, but probably others will have different requirements.

Please note: I did not write any unit-tests for my solution yet, as I was not sure, if you like my approach. Please feel free to add some or let me know, if you want me to do it.

With these small changes we are almost completely ready to go with our project - so thanks for the great work!!! It really helps a lot!! 

BTW: I filed a small error concerning duplicate attribute-definitions. I guess you are a little faster spotting the exact point to fix it.

I will stay in touch with you!

Thomas

Von meinem iPad gesendet

vojtechhabarta:
#60 finished this work.

PR #63, opened 2016-05-17
Add enum values bean mapping

cghislai:
This change introduces a new option to allow the creation of an interface for enum constants. This is welcomed in case fields are declared inside the java enum.
As the type name must be suffixed, another option to specify that suffix has been added.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -40,6 +40,8 @@ public class Settings {
     public List<EmitterExtension> extensions = new ArrayList<>();
     public List<Class<? extends Annotation>> optionalAnnotations = new ArrayList<>();
     public boolean experimentalInlineEnums = false;
+    public boolean enumValueModel = false;
+    public String addEnumValueSuffix = "Value";
 
 
     public void loadCustomTypeProcessor(ClassLoader classLoader, String customTypeProcessor) {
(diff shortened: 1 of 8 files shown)


vojtechhabarta:
If I understand it correctly this change adds possibility to generate TypeScript interface for data produced by custom Jackson `JsonSerializer`. I think this should not be part of the core because people can have many different custom `JsonSerializer`s. Intended way for this use case is to write interface for custom enum format by hand and use `customTypeMappings` parameter (together with `referencedFiles` or `importDeclarations`) as described on [Type Mapping](../wiki/Type-Mapping) Wiki page.

What is your use case? Do you use many enums with this custom format? If you still want to generate the interfaces automatically you can implement `EmitterExtension`. For example see [EnumConstantsExtension](../blob/master/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/EnumConstantsExtension.java).

cghislai:
Indeed this was not intended for direct mapping to default Jackson serialized enum.
I wanted to make the enum fields available in the typescript model and fetch them once on application startup for each enum constants. I want to avoid the need to maintain these static properties on both sides, java and typescript.

After a quick look to custom type mappings and EmitterExtension, it seems to me that the java enums must still be parsed in a model including the fields.

I will suggest another push request probably including an EmitterExtension.

vojtechhabarta:
Yes, Java enum fields still need to be read using reflections but this can be done in the extension itself. It's not needed to modify the model. You can get enum class using `TsEnumModel.getOrigin()` method.

And I think it is better to do it in the extension because it is really specific. For example you are using fields (including private) and somebody else might want to use getters (public) in another extension.

PR #64, opened 2016-05-22
Add annotationFilters and displaySerializerWarning settings

ZeroErrors:
Add the ability to do two things:
- Filter the methods generated by requiring an annotation to be present.
- Add the ability to disable the "Warning: Unknown serializer" messages.

Method filtering does not yet happen in the JaxrsApplicationScanner but should probably be added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ release.properties
 **.classpath
 **.project
 **.settings
+
+# JetBrain IntelliJ
+.idea/
+**.iml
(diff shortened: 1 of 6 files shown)


vojtechhabarta:
Thanks for these features.
I would just like to rename `annotationFilters` to something like `includePropertyAnnotations`. I think sooner or later somebody will want opposite of this feature which we can then name `excludePropertyAnnotations`. And I would keep this parameter specific to properties.

BTW are you using some public annotations or do you have your custom annotation?

ZeroErrors:
Yep, makes sense.

The project I'm working on uses a custom annotation.

Issue #77, opened 2016-08-19
Enum support


rgevrek:
**Enum**s are turned into **type**s, but they could be turned into **enum**s in TypeScript, as in this following example:

`export enum MyEnum { el1, elem2 }`

instead of:

`export type MyEnum = "el1" | "elem2";`

vojtechhabarta:
Main reason for this is that Java enum values are sent as strings in JSON. See #65 or other issues for more info.

rgevrek:
Thank you for your answer, I've added the support for enums in this [pull request](https://github.com/vojtechhabarta/typescript-generator/pull/78)

hypery2k:
is it possible to release the enum extension ?

vojtechhabarta:
I don't want to add this feature as core functionality for two reasons:
1. As I mentioned TypeScript enums are **number** based but Java enums are sent as **strings** in JSON.
2. TypeScript enums (even with `const` keyword) don't work well in declaration files (`.d.ts`).

Here is example for point 2:
_Declaration file with enum_
```
// enums.d.ts
export const enum Direction {
     North, East, South, West
}
```
_Implementation file which uses the enum_
```
// test.ts
import * as enums from "./enums";
const value: enums.Direction = enums.Direction.East;
console.log(value);
```

If you run TypeScript compiler it successfully generates `test.js` file but this file fails in runtime (using `node test.js`). I can describe this in more details if you want or you can try it and check what TypeScript generates and why it fails.

@rgevrek If you really want to have this functionality despite mentioned drawbacks you can change your PR to contain just an extension (and not `typescriptEnums` core option).

I would also like to hear from you guys the motivation for this feature. Maybe I will convince you that you don't need it ;-)

ollwenjones:
If all one does is pass "ENUM_VAUE" strings around the client, and all you want is typescript to complain if you misspell something, then 
```TypeScript
type MyEnum = "BOB" | "SALLY" | "JIM";
```
is fine, but if you want to do some normal enum opration, i.e. compare those things as enums, for instance:
```TypeScript
    let person = "BOB";
    if (person === MyEnum.BOB) {
```
you can't do that.

would be nice if it was an object that presented itself as if it were a Java enum:
```JavaScript
const MyEnum {
  BOB: 'BOB', //etc.
...
```
but that doesn't work in a `.d.ts`

vojtechhabarta:
[...]

PR #78, opened 2016-08-19
Enum support #77 added

rgevrek:
With this change you can get following enum structure:
`enum MyEnum { el1, elem2 }`

by adding following configuration:
`<typescriptEnums>true</typescriptEnums>`

or by adding this extension:
```
<extensions>
<extension>cz.habarta.typescript.generator.ext.EnumExtension</extension>
</extensions>
```

### DIFF
--- a/sample-maven/pom.xml
+++ b/sample-maven/pom.xml
@@ -21,7 +21,7 @@
             <plugin>
                 <groupId>cz.habarta.typescript-generator</groupId>
                 <artifactId>typescript-generator-maven-plugin</artifactId>
-                <version>1.7-SNAPSHOT</version>
+                <version>1.10-SNAPSHOT</version>
                 <executions>
                     <execution>
                         <id>generate</id>
@@ -35,6 +35,7 @@
                             </classes>
                             <outputFile>target/sample.d.ts</outputFile>
                             <outputKind>module</outputKind>
+                            <typescriptEnums>true</typescriptEnums>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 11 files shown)


rgevrek:
const prefix needed for enum

vojtechhabarta:
@rgevrek I will add Java enum to TypeScript enum feature but I want to implement it somewhat differently:
- use `Emitter.emitNumberEnums` for generating all TypeScript `enum`s
- use `TsEnumModel` as TypeScript enum representation (not `TsAliasModel` which represents [Type Aliases](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases))
- not adding new `TsType` - `TsType` represents TypeScript type, something which can be used in type position (for example `let a: <TsType>`) - `CommaSeparatedType` doesn't represent TypeScript type in this sense
- unify configuration switch for this feature with inline enums feature

I know that typescript-generator code is not well documented so it is not easy to know how to make changes.

Anyway I highly appreciate your interest and contribution to typescript-generator. Thanks

rgevrek:
@vojtechhabarta thank you for your help. I think it's even better if you do the implementation, too, to have the changes in right place.

In .ts files following solution is working well, but as you mentioned this is not possible for .d.ts files:

````
type MyEnum = "BOB" | "SALLY" | "JIM";
const MyEnum = {
    BOB: <MyEnum>"BOB",
    SALLY: <MyEnum>"SALLY",
    JIM: <MyEnum>"JIM",
}
```

But with this it's type safe in development, and you get a perfect match with Jackson JSON parser

PR #83, opened 2016-09-28
Extension for enums without const keyword

vojtechhabarta:
Based on discussion in #77 this PR implements extension which from Java enums generates TypeScript enums without `const` keyword.
This extension is implemented in `cz.habarta.typescript.generator.ext.NonConstEnumsExtension` class.

By default typescript-generator generates string literal unions like this
``` typescript
type Direction = "Left" | "Right";
```
which exactly describes JSON data and works well, especially in TypeScript 2.0.

In some cases it might be useful to generate non-const enums using this extension (instead of string unions) like this:
``` typescript
enum Direction {
    Left,
    Right,
}
```

These enums only work in implementation files (`.ts`) because they are real objects that exist at runtime, it is not sufficient to put them into declaration file (`.d.ts`).

> Note: Java enums are sent in JSON as **string** values while TypeScript enums are **number** based. So this extension should only be used if there is a good reason.

More information about TypeScript enums can be found in Handbook - https://www.typescriptlang.org/docs/handbook/enums.html.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -101,7 +101,7 @@ public class Settings {
         }
         if (outputFileType != TypeScriptFileType.implementationFile) {
             for (EmitterExtension emitterExtension : extensions) {
-                if (emitterExtension.generatesRuntimeCode()) {
+                if (emitterExtension.getFeatures().generatesRuntimeCode) {
                     throw new RuntimeException(String.format("Extension '%s' generates runtime code but 'outputFileType' parameter is not set to 'implementationFile'.",
                             emitterExtension.getClass().getSimpleName()));
                 }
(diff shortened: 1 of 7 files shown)


vavasilev:
Hello,

I tested the fix - it duplicates the enums - both const and non-const ones are generated

vojtechhabarta:
I guess you configured `NonConstEnumsExtension` and also set `mapEnum` to `asNumberBasedEnum`.
I think I fixed this case, only non-const enums should be generated.

vavasilev:
Yes, this is my configuration:

<configuration>
                     <classes>
                        ...
                     </classes>
                     <classPatterns>
                        ...
                     </classPatterns>
                     <excludeClasses>
                        <excludeClass>java.io.Serializable</excludeClass>
                     </excludeClasses>
                     <extensions>
                        <extension>cz.habarta.typescript.generator.ext.NonConstEnumsExtension</extension>
                     </extensions>
                     <outputFileType>implementationFile</outputFileType>
                     <outputFile>../src/app/modules/global/model/emitted/emitted.ts</outputFile>
                     <outputKind>module</outputKind>
                     <mapEnum>asNumberBasedEnum</mapEnum>
                  </configuration>

vojtechhabarta:
You can remove `<mapEnum>asNumberBasedEnum</mapEnum>` if you use `NonConstEnumsExtension`.
(`mapEnum` parameter doesn't have any effect when the extension is used)

vavasilev:
Removing mapEnum worked! Thanks!

One more question. When will all the latest stuff be available on maven repository?

PR #87, opened 2016-10-17
fixes #86 - new emitter extension to generate bean property paths.

### LINKED ISSUE #86: make more extendable by allowing to plug-in custom emitters?
A colleague of mine opened https://github.com/vojtechhabarta/typescript-generator/issues/22 some time ago. We didn't fix that issue yet but now we're far enough that it's becoming a bigger hindrance and it's again on our radar.
You mentioned solutions (1), (2) and (3).
I now think only (3) is realistic for us, because besides the ng-grid use-case we also internally developed some filter mechanism where the client-side sends to the server a filter specification. That specification also refers to fields in the DTOs in the same way as ng-grid, and we don't intend to change that (it's a different problem than for ng-grid, because our specification is going over the wire to the server).

Also (2) is interesting, but incomplete, for instance we often have specifications like "field1.field2" (with field2 most of the time belongs to another class than field1).

So the ideal solution would be to generate some TS code which would allow us to type something like:

ClassName.field1.get() => "field1"
ClassName.field1.field2.get() => "field1.field2"

I've looked a bit at the code, and I think it would be possible for us to effectively fork typescript-generator-maven-plugin, and when it calls typescript-generator-core, we would paste-in that definition:
https://github.com/vojtechhabarta/typescript-generator/blob/a38a6f4deb02e1b8892196371d444abc33951cae/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TypeScriptGenerator.java#L48-L52

and use the TsModel to generate the TS code that we want, besides calling typescript-generator-core's generator to generate the TSD as usual (I didn't check yet if enough of the classes from the core are really public, but I think they are).

But I don't like forking typescript-generator-maven-plugin, especially because we would probably put our fork on maven-central. So I'm wondering if there is maybe a way to get to the TsModel already now, and achieve our purpose, or maybe such an extension mechanism could be added?

emmanueltouzery:
This is is the emitter extension that we discussed. I've also made it so that inherited fields are generated.
I could make the test compare exactly the output, if you prefer -- I just made it match the number of lines and added some minimal sanity check about strings it should contain.

PS: note that the "| undefined" compiles only with typescript2, but I guess it's a feature -- typescript2 is the future and even the present.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/BeanPropertyPathExtension.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/BeanPropertyPathExtension.java
@@ -37,8 +37,11 @@ public class BeanPropertyPathExtension extends EmitterExtension {
     @Override
     public void emitElements(Writer writer, Settings settings, boolean exportKeyword, TsModel model) {
         emitFieldsClass(writer, settings);
+
+        Set<TsBeanModel> emittedBeans = new HashSet<>();
         for (TsBeanModel bean : model.getBeans()) {
-            writeBeanFieldSpecs(writer, settings, model, bean);
+            emittedBeans.addAll(
+                writeBeanAndParentsFieldSpecs(writer, settings, model, emittedBeans, bean));
         }
         for (TsBeanModel bean : model.getBeans()) {
             createBeanFieldConstant(writer, bean);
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
I tried the extension and it generates duplicate fields in one special case.
If a class contains some field which is also inherited then the extension generates definition for this field multiple times. This happens in case of tagged unions. See https://github.com/vojtechhabarta/typescript-generator/pull/81. In this example `kind` field is generated twice in derived classes.

I think ideally derived "fields" class could extend base "fields" class instead of "inlining" inherited fields. So in our example this could be generated:
```` typescript
class RectangleFields extends ShapeFields {
    constructor(parent?: Fields, name?: string) { super(parent, name); }
    kind = new Fields(this, "kind");
    width = new Fields(this, "width");
    height = new Fields(this, "height");
}
```
This would fix the problem and as a bonus it wouldn't duplicate fields from base classes in derived classes.
But the tricky part here is that base "fields" classes must be generated before derived classes.

What do you think?

emmanueltouzery:
good point! Made a second commit with the suggested changes. Maybe squash them if you pull.

emmanueltouzery:
btw with this change our generated TS file reduced from 1100 LOC to 1055 due to non-duplication of the fields. not that much but less code is always nice.

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/BeanPropertyPathExtension.java
```diff
@@ -0,0 +1,136 @@
...
+        Writer writer, Settings settings, TsModel model, TsBeanModel bean,
+        TsPropertyModel property) {
+        TsBeanModel fieldBeanModel = getBeanModelByType(model, property.getTsType());
+        String fieldClassName = fieldBeanModel != null ? getBeanModelClassName(fieldBeanModel) : "";
+        writer.writeIndentedLine(
+            settings.indentString + property.getName() + " = new " + fieldClassName + "Fields(this, \"" + property.getName() + "\");");
+    }
+
+    private static void createBeanFieldConstant(Writer writer, TsBeanModel bean) {
+        writer.writeIndentedLine("export const " + getBeanModelClassName(bean) + " = new " + getBeanModelClassName(bean) + "Fields();");
```
emmanueltouzery: Done. I thought you need `export` when generating .ts as opposed to .d.ts -- for us here without export, the typescript code doesn't build.

emmanueltouzery:
forget what i said about not building without the export keyword, I just had a bug :-)

vojtechhabarta:
Thank you very much for your PR. I merged it and squashed your commits.

vojtechhabarta:
released: https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.11.233

PR #88, opened 2016-10-26
Added an option allowing generating the Java types as 'classes' 

hsq125:
Hi Vojtěch!

Thanks for this great plugin. I'm using it to generate declaration for Angular, and need to have 'class' instead of 'interface'. 

I added an option **classType** to select between _class_ or _interface_.

Feel free to merge into your code base.

Best,
O.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ClassType.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ClassType.java
@@ -1,0 +1,16 @@
+package cz.habarta.typescript.generator;
+
+public enum ClassType {
+
+    asInterface("interface"), asClass("class");
+
+    public final String keyword;
+
+    ClassType(String keyword) {
+        this.keyword = keyword;
+    }
+
+    public String getKeyword() {
+        return keyword;
+    }
+}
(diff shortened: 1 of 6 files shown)


vojtechhabarta:
Hi @hsq125, thanks for your interest in this tool.

Do you want to generate classes in declaration file (`.d.ts`) or implementation file (`.ts`)?

Typescript-generator produces by default declaration file while classes are runtime objects. If we generated classes in declaration file it would mean that those classes actually exist somewhere else and generated file would not be usable alone.

So I think we want to generate classes only in implementation files. For features that are not intended for declaration files typescript-generator has concept or extensions. There is for example extension which generates constants for enum values `cz.habarta.typescript.generator.ext.EnumConstantsExtension`.

Would you be interested in implementing this feature as an extension? If yes, I will add possibility for plugins to indicate that they are overriding generated interfaces (add `EmitterExtensionFeatures.overridesInterfaces`).

hsq125:
Hi Vojtěch,

Sorry for late reply. Indeed, we want to generate classes only in implementation file (`.ts`). I'm ok to implement that as an extension, so please open the possibility.

Thanks,
O.

vojtechhabarta:
@hsq125 sorry for late response.
Finally I decided to add this feature as core functionality but I wanted to implement it somewhat differently, see #97 for details.
Sorry for not using your code but I hope this feature will meet your needs.

hsq125:
Hi Vojtěch,

That's great! I managed to configure the plugin to get exactly what I need! Maybe you are interested to state somewhere that your tool can be configured as follow to generate .ts file for Angular object.

```xml
<configuration>
    <classPatterns>
          <pattern>td.web.dto.*</pattern>
    </classPatterns>
     <outputFileType>implementationFile</outputFileType>
     <outputFile>target/model.ts</outputFile>
     <outputKind>module</outputKind>
     <mapClasses>asClasses</mapClasses>
</configuration>
```
Thanks again!
Octave

PR #93, opened 2016-11-18
Support for Java 8 Optional

vojtechhabarta:
With this PR typescript-generator recognizes `java.util.Optional` types and maps `Optional<T>` in Java to TypeScript as it was just `T`.

For example for this Java class:

``` java
public class Person {
    public String name;
    public Optional<String> email;
}
```

it generates following TypeScript declaration:

``` typescript
interface Person {
    name: string;
    email: string;
}
```

**Use case** for this feature is to allow Java code to detect whether some JSON property was sent with `null` value or it wasn't send at all (`undefined` in JavaScript).

Normally both cases are represented as `null` value in Java but Jackson2 can be configured to distinguish these two cases using Java 8 `Optional` types:

``` java
ObjectMapper objectMapper = new ObjectMapper()
        .registerModule(new Jdk8Module())
        .setSerializationInclusion(JsonInclude.Include.NON_NULL);
```

To use `Jdk8Module` it is needed to add dependency on `com.fasterxml.jackson.datatype:jackson-datatype-jdk8` artifact.

Jackson2 then serializes/deserializes `Optional` properties as follows:

Java property value | JSON property value
--- | ---
`Optional.of(value)` | `value`
`Optional.empty()` | `null`
`null` | `undefined` (property not sent)

### DIFF
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -24,6 +24,11 @@
             <artifactId>jackson-jaxrs-json-provider</artifactId>
             <version>2.6.3</version>
         </dependency>
+<!--        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jdk8</artifactId>
+            <version>2.6.3</version>
+        </dependency>-->
         <dependency>
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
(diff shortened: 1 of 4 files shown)


cluelessjoe:
Is there a way to use this feature from the maven plugin?

Thanks in advance

vojtechhabarta:
@cluelessjoe yes, this support for `Optional` works regardless how typescript-generator is invoked (using typescript-generator-maven-plugin, typescript-generator-gradle-plugin or directly from Java). No need to configure anything.

cluelessjoe:
Thanks for this quick answer. Using the plugin from your example: 
```
<groupId>cz.habarta.typescript-generator</groupId>
<artifactId>typescript-generator-maven-plugin</artifactId>
<version>1.25.322</version>
```
this didn't work: no ? next to the TS's type. Or I'm missing something, quite common too ;)

thanks a lot for your work BTW :)

vojtechhabarta:
Currently typescript-generator doesn't generate `?` for `Optional` types. It just "unwraps" the type.
I will change it.

cluelessjoe:
That would be nice to know on the TS side that's some values are optional :)

PR #94, opened 2016-11-18
bean property path extension: bug if the DTO had a field named 'name' or 'parent'

emmanueltouzery:
with bug #86 we added an emitter extension to generate type-safe property accessors. We now found a little bug in it here. Basically the Field class for every DTO inherits from the base `Fields` class.

And the `Fields` class has fields `name` and `parent`. Now, if any DTO has a field named `name` or `parent`, it will shadow the field from the `Fields` class and cause all kinds of problems :-)

So I renamed the fields, prepending $$. Since $ is an illegal character for java variable names (especially at the start), no collision is possible after this change.


vojtechhabarta:
Thanks for this fix.

PR #95, opened 2016-11-18
Fix NPE in JaxrsApplicationScanner#isExcluded

ollwenjones:
I was getting an NPE in the for loop in `JaxrsApplicationScanner.isExcluded` at `standardEntityClass.isAssignableFrom(cls)`

Is there ever a case where `cls` is null, and we want to _include_ it?

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/JaxrsApplicationScanner.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/JaxrsApplicationScanner.java
@@ -128,7 +128,7 @@ public class JaxrsApplicationScanner {
     private boolean isExcluded(Type type) {
         final Class<?> cls = Utils.getRawClassOrNull(type);
         if (cls == null) {
-            return true;
+            return false;
         }
         if (isClassNameExcluded != null && isClassNameExcluded.test(cls.getName())) {
             return true;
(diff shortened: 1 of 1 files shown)


vojtechhabarta:
Good quiz ;-)
Actually I think I found some case: `List<String>[]` (generic array). `Utils.getRawClassOrNull()` returns `null` for such type but that type should not be excluded. Typescript-generator currently doesn't support generic arrays but it could.

I would return `false` when `cls` is `null`. Does this work for you?

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/JaxrsApplicationScanner.java
```diff
@@ -127,12 +127,16 @@ private void foundType(Type type, Class<?> usedInClass, String usedInMember) {
 
     private boolean isExcluded(Type type) {
         final Class<?> cls = Utils.getRawClassOrNull(type);
-        if (cls != null && isClassNameExcluded != null && isClassNameExcluded.test(cls.getName())) {
+        if (cls == null) {
             return true;
```
vojtechhabarta: Here we should return `false`, I think.

ollwenjones:
You understand more of what the output of that is producing than I do, so I would take your word for it, except changing that to `false` fails `InputTest` for me. My line of thinking was a null type would always be an empty interface that wouldn't enforce anything in TypeScript, even if it corresponded to something in the JSON. I could be wrong though.

ollwenjones:
passed after logging to see what was what... mysterious.

ollwenjones:
Looks like the results are the same for my use case, so if you think it should be false, I'll take your word. My main hurdle was the build failing on the NPEs

vojtechhabarta:
Thanks for this fix.

PR #100, opened 2016-12-02
bean property path extension: link between objects in more cases

emmanueltouzery:
we've found another thing to improve in the bean property path extension...

if we have:

class CustomerDto { name: string; }
class UserDto { customer?: CustomerDto | null | undefined; }

then it would not detect the link between classes and you could not get the following path =>

"UserDto.customer.name.get()"

same if UserDto had CustomerDto[]. So we now support these situations so that the link is established despite these issues.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/BeanPropertyPathExtension.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/BeanPropertyPathExtension.java
@@ -101,9 +101,52 @@ public class BeanPropertyPathExtension extends EmitterExtension {
         return emittedBeans;
     }
 
+    /**
+     * is this type an 'original' TS type, or a contextual information?
+     * null, undefined and optional info are not original types, everything
+     * else is original
+     */
+    private static boolean isOriginalTsType(TsType type) {
+        if (type instanceof TsType.BasicType) {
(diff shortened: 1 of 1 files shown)


vojtechhabarta:
@emmanueltouzery few minutes ago I merged some changes in model (#101) which might affect your PR. I will merge your PR, could you please check if everything works as expected then?

emmanueltouzery:
@vojtechhabarta I've just rebuilt our app with the latest master and I don't see any issues. Looks good to me! [i'll still stick with the released version for our PROD release ;-) ]

vojtechhabarta:
@emmanueltouzery thanks 👍

emmanueltouzery:
@vojtechhabarta  sorry I spoke too fast. I tried the latest code with my own extension code. That works. But removing the generics magic as you did in the extension code does break it! With the generics code it's fine. I think it should be reverted. It appears it's still needed

vojtechhabarta:
I think problem is somewhere else. `TsBeanModel.getName()` now returns `Symbol` (not `TsType`) so comparing it with `originalType` doesn't make sense (https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/BeanPropertyPathExtension.java#L145). You can narrow `originalType` to `ReferenceType`, get the `Symbol` and compare it.

Sorry for breaking your code, the test didn't catch this.

PR #102, opened 2016-12-02
don't make a stack overflow when there is a type referring to itself

emmanueltouzery:
sorry, my previous PR was maybe a little too eager. I now noticed this issue, because we now will also link DTOs through List references, like so:

class X { myField: List<X> }

and if we have that, we get a stack overflow in the generated typescript. So with this change I just don't generate the link anymore in case it's a self-reference. Workaround, but it works.

PS: I don't know if you saw that I think we need the generics code that you removed on master => https://github.com/vojtechhabarta/typescript-generator/pull/100#issuecomment-264481195


emmanueltouzery:
regarding the issue of the generics code & your recent changes I think we need better tests for this feature. For instance we could have the expected output in a tests resource file & we could compare the output.

PR #105, opened 2016-12-17
bean property path extension: export the Fields class

emmanueltouzery:
one more change to the extension... exporting the `Fields` class turns out to be useful.

We've had the situation for instance where one DTO was something like that:

```java
public class ExportDto<T> {
  T dto;
  boolean isOk;
  ...
}
```

clearly it's going to be tricky to access it with the bean property path mechanism. What we did is:

```typescript
const getExported = (field: Fields) => ExportDto.dto.get() + "." + field.get();
...
field: getExported(MyProperDto.myProperField);
```

turns out quite clean, and as type-safe as possible. But if we don't export the `Fields` type, `getExported` can only take `any` and it's less safe.


vojtechhabarta:
@emmanueltouzery this is now part of [1.15.265](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.15.265) version. Sorry for late release.

emmanueltouzery:
Just taking the opportunity to thank you for your amazing work on this package!

Issue #115, opened 2017-02-06
Generating interface for JAX-RS application


vojtechhabarta:
It is now possible to generate TypeScript interface for JAX-RS application.

For example for this REST resource class:
``` java
@Path("people/{personId}")
public class PersonResource {

    @PathParam("personId")
    protected long personId;

    @GET
    public Person getPerson() {
        return null;
    }

    @GET
    @Path("address/{addressId}")
    public Address getAddress(@PathParam("addressId") long addressId) {
        return null;
    }

}
```

typescript-generator can produce this interface:

``` typescript
interface RestApplication {
    /**
     * HTTP GET /people/{personId}
     */
    getPerson(personId: number): RestResponse<Person>;
    /**
     * HTTP GET /people/{personId}/address/{addressId}
     */
    getAddress(personId: number, addressId: number): RestResponse<Address>;
}
```

Above mentioned example is simple so try this feauture yourself on your application, ask questions and provide some feadback.

You can enable this feaure using `experimentalJaxrsApplicationInterface` parameter.

_Next step will be to generate implementation._

vojtechhabarta:
Implementation is part of [v1.18.288](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.18.288) release.

doblak:
Great feature, thanks.

But when using client application interface generation with dozens of services all the REST methods get emitted to the same interface resulting in one big mess and name collisions. It's somehow related to #99 but in context of client interface.

It would make sense to me emitting one interface per resource class, e.g. `interface PersonResorce` instead of `interface RestApplication`

Do you have any suggestions? Or is there any other way to make this work?

vojtechhabarta:
To me generate one TypeScript interface per one Java resource class is too **fine grained**, resources usually only have few operations.

But you are right that single client interface is too **coarse grained**, there are usually many name conflicts.

Ideally I would like to have operations somehow grouped to several interfaces. But this would probably require to have some non-intuitive configuration. What do you think?

Currently workaround is to run typescript-generator multiple times and configure each run to process different resource(s). In this case some data interfaces could be generated multiple times but this should not be a problem since TypeScript has structural typing.

doblak:
In our case it feels natural to have one TS interface per one Java resource. Java resources in our case are organized as services usually containing ~5-20 calls.

I'd opt to offering 2 default options - single interface and 1 interface per 1 resource. 

[...]

PR #126, opened 2017-03-20
JAX-RS namespacing options

vojtechhabarta:
When generating declarations for JAX-RS application typescript-generator created single declaration with all REST operations. This is convenient but often results in many name conflicts. Based on feedback in #115 this PR adds support for splitting operations into multiple declarations. This provides more namespaces for operations and reduces name conflicts.

This change adds `jaxrsNamespacing` parameter. Possible values are:
- `singleObject`: original behavior, currently default
- `perResource`: one declaration per root resource
- `byAnnotation`: operations grouped by annotation specified using `jaxrsNamespacingAnnotation` parameter, for example: `io.swagger.annotations.Api#value`

Description on [JAX-RS Application Wiki page](https://github.com/vojtechhabarta/typescript-generator/wiki/JAX-RS-Application#options-for-grouping-operations-namespacing).


doblak:
Tested `perResource` and it works great. I modified the AxiosClientExtension and templates a bit so they fit our specific workflow.

Thanks again for your effort!

PR #127, opened 2017-03-21
Adding optional support for joda-time's DateTime class

jthoenes:
- we check, if JavaTime is available in class path
- if it is, we handle it exactly like java.util.Date

### DIFF
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -73,7 +73,13 @@
             <artifactId>swagger-annotations</artifactId>
             <version>1.5.10</version>
             <scope>test</scope>
-        </dependency>        
+        </dependency>
+        <dependency>
+            <groupId>joda-time</groupId>
+            <artifactId>joda-time</artifactId>
+            <version>2.9.7</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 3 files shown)


vojtechhabarta:
@jthoenes thanks for your PR.

I don't want to process types from external libraries in `DefaultTypeProcessor`. But in this case it is just one class (`org.joda.time.DateTime`) so lets leave it in `DefaultTypeProcessor` for now and it can be refactored into some new `TypeProcessor` later.

vojtechhabarta:
Part of [1.22.307](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.22.307) release.

vojtechhabarta:
@jthoenes I removed support for `org.joda.time.DateTime` type in [v1.28.343](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.28.343) because I want to minimize specific library handling in typescript-generator core. Sorry for that.

Instead I added support for `java.time` date types (similar to joda types but part of Java 8). All types that implement `java.time.temporal.Temporal` are now considered to be date types. Jackson2 supports these classes using `com.fasterxml.jackson.datatype:jackson-datatype-jsr310` module which can be added to Jackson2 using
``` java
new ObjectMapper().registerModule(new JavaTimeModule())
// or
new ObjectMapper().findAndRegisterModules()
```

Joda's `org.joda.time.DateTime` still can be used. It can be mapped for example to `string` using `customTypeMappings` parameter. Here is Maven configuration sample:
``` xml
<configuration>
    <customTypeMappings>
        <mapping>org.joda.time.DateTime:string</mapping>
    </customTypeMappings>
    ...
</configuration>
```

PR #151, opened 2017-06-24
Sort the discriminants by the tagged unions of the first matching parent

glentakahashi:
@vojtechhabarta So, this is an alternative to the other PR that just merged. I meant to say [DONT MERGE] because I wanted to open it up for discussion, but forgot to add that.

This one seems a little heavy handed, but I wanted to throw it out there. Basically, this finds the first matching parent and uses the tagged unions on those to sort the discriminants. This means that all the orderings will be the same across the types.

The one worry that I had was about my use of getParentAndInterfaces(), because I wasn't sure if it was possible for a class to implement multiple classes which have taggedUnions. E.g. for two interfaces with @JsonTypeInfo, I don't think it's possible for a class to implement both, is it? If that were the case, you wouldn't be able to guarantee non-clashing type names.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
@@ -139,10 +139,19 @@ public class ModelCompiler {
             }
         }
         final List<TsPropertyModel> properties = processProperties(symbolTable, model, bean, "", "");
+        final List<Class<?>> parentTaggedUnionClasses = getParentTaggedUnionClasses(bean, model);
 
         if (bean.getDiscriminantProperty() != null && !containsProperty(properties, bean.getDiscriminantProperty())) {
             final List<BeanModel> selfAndDescendants = getSelfAndDescendants(bean, children);
             final List<TsType.StringLiteralType> literals = new ArrayList<>();
+            if (parentTaggedUnionClasses != null) {
+                Collections.sort(selfAndDescendants, new Comparator<BeanModel>() {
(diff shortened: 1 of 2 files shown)


glentakahashi:
If you like this implementation, I would prefer to revert my other PR and merge this instead.

vojtechhabarta:
@glentakahashi I tried this PR and it doesn't work in some special case.
I tried something like this:
``` java
class A {}
class B extends A {}
class C1 extends B {}
class C2 extends B {}
```
where class A has `@JsonSubTypes` annotation with B, C1, C2.

I can prepare test for this case if you want.

glentakahashi:
Hmm. I just tried that case and it seemed to work for me. Could I see the
test case that fails? No rush on this by the way!

Also in the meantime, do you want to maybe revert my PR for the other one,
and we can merge back in later if it's necessary?

glentakahashi:
Oh wait I see where it failed, I rearranged the ordering and that's not
working. Lemme see if I can find out why.

Glen

glentakahashi:
Ah @vojtechhabarta i figured it out. Essentially, when beans are created you actually pass in the .type most of the time instead of .class. For interfaces/abstract classes, getClass will return Class instead of what you want, so I use Type instead. I was thinking about just refactoring bean Origin to be Type everywhere, but wanted to run that by you first.

It gets kind of ugly though, because for some reason Object is a parent of Object, so i have to add an explicit check to not recurse on that class.

Up to you on which method of discriminant ordering you want to include. I have no real preference one way or the other

glentakahashi:
Have to close and recreate a PR because github wasn't updating the PR for some reason

PR #153, opened 2017-06-26
Glen/fix interface inheritance types

glentakahashi:
@sayresRT this should fix your bug and add in your test

The bug was previously that getInheritedProperties would not handle duplicates, and if two parents differed it was just the last one that won out. If the last one happened to equal your current property (as it happened to do in your use case), it would think they're compatible and remove just that one property.

Essentially what I did was rewrite the "getInheritedProperties" method to create a new "collision type" if two parents have differing types for a property. this means that the equality check when removing inherited props will fail and the class will include the property itself.

If you're two degrees away (i.e. your parent inherits incompatible properties), splitting the function into two recursive functions causes single classes to override their parents if they collide, which gives us the functionality that we want.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
@@ -237,6 +237,26 @@ public abstract class TsType {
 
     }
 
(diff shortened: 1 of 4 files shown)


glentakahashi:
Closed because irrelevant now

Issue #168, opened 2017-09-07
question: generating equals & hashcode


emmanueltouzery:
the [immutablejs](https://facebook.github.io/immutable-js/) framework for instance allows you to have `Set` of items, but so that no two duplicate items are in fact added in the Set, the user must provide equality semantics for the items.

In particular, the user must provide two methods for items:
https://facebook.github.io/immutable-js/docs/#/ValueObject

`equals` and `hashCode`. What I would like to have is getting typescript-generator to optionally generate such functions for the mapped DTOs. So, we generate classes, not interfaces, and add these methods.

Now clearly there are multiple possible equality semantics. Sometimes we may want to compare only by the ID field. And sometimes we want to recursively compare every field. Also a function must be written to calculate the hash of basic types, such as strings. So I think the full logic for that doesn't belong in typescript-generator. But I'd love to have an extension point, a possibility to write a plugin to do that. Do you think it's possible?

For our project, ideally, I could see toplevel annotations on DTOs like `@EqualityById` (must have an `id` field) and `@EqualityByValue` or something like that. Then our plugin affects the code generation. Adds these extra methods.

There is a side-task too. We currently generate a `.ts` already, not a `.d.ts`, but we generate interfaces for our types. Generating classes would bring in some bloat. It's not that often that we need Sets.. So ideally we would still generate interfaces, and generate classes only when one of these annotations, EqualityById, or EqualityByValue is present (it would be a typescript compile error down the line if a class has the annotation but has a field with a non-primitive type without that annotation, no need to handle that ourselves). I was looking at `ModelCompiler.processBean`. Even though nominally the setting to generate a class or an interface is global for the whole system, in the code, it's actually per-class. So I think there an extension could also be possible. That typescript-generator could invoke a plugin in `processBean` and the plugin would say whether, for that particular class, it should generate a class or an interface.

What do you think? Do you think these features are possible without complicating typescript-generator too much?

vojtechhabarta:
Generate mix of classes and interfaces based on some annotations could be useful feature in general.

What do you expect to be generated? Something like this:
``` typescript
// annotation in Java was @EqualityById
class Company {
    id: string;
    name: string;
    hashCode(): number {
        return hashCodeHelper(this.id);
    }
    equals(other: any): boolean {
        return equalsHelper(this.id, other.id);
    }
}

// annotation in Java was @EqualityByValue
class Person {
    name: string;
    age: number;
    hashCode(): number {
[...]

PR #170, opened 2017-09-10
ability to enforce that a class is generated as a class not as interface

emmanueltouzery:
ability to enforce that a class is generated as a class not as interface, even if we generate interfaces according to the global settings, see #168.

this is the first step for #168, the second being the ability to add extra contents in generated classes through extensions.


vojtechhabarta:
This is now possible using `mapClassesAsClassesPatterns` parameter which is based on class names (not on annotations). I hope this should be sufficient but we can of course discuss it.

PR #171, opened 2017-09-10
allow extensions to add extra code in generated classes

emmanueltouzery:
this is related to #168 -- add ability for extensions to add code in generated classes.

I have experimented with an extension to add hashcode/equals to classes with the right annotations, you can see a gist there:
https://gist.github.com/emmanueltouzery/200edb5d4611441584c501161fd2e622

It looks good in the TS we generate. However before I test the whole chain on our side, I probably need the feature we mentioned, to instantiate the classes in the JAX-RS client; but in theory, I think it should work!

PS: might be nice to add to the writer a function to add some source without trailing carriage return, it's a bit annoying sometimes to have to have the CR always.


emmanueltouzery:
so this PR may be obsolete since you suggested the alternative solution of overwriting the prototype of the class.

vojtechhabarta:
Closing issues/PRs after [v2 release](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.0.400). We can reopen if needed.

Issue #172, opened 2017-09-11
Empty/default jackson @JsonTypeName isn't handled correctly


Yona-Appletree:
According to the JavaDoc on Jackson's @JsonTypeName, if it's value is left blank, it defaults to the non-qualified class name of the annotated class. typescript-generator, however, actually emits the empty string, when it should emit the simple class name.

vojtechhabarta:
Resolved in #178. Released [v1.29.355](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.29.355).

PR #176, opened 2017-09-18
Fix for #172 - Handling empty @JsonTypeName() correctly.

Yona-Appletree:
Fixes Issue #172 by checking for empty values in @JsonTypeName and using the default logic when that is the case.


Yona-Appletree:
Closed to create PRs for each fix independently

PR #177, opened 2017-09-18
Bug 172 - Handling empty @JsonTypeName

Yona-Appletree:
Adds logic to check for empty value of `@JsonTypeName` and defaults to existing logic, as indicated by the JavaDoc on `@JsonTypeName`


Yona-Appletree:
Closing because I accidentally pulled in a feature branch from origin.

PR #178, opened 2017-09-18
Fix for #172 - Handling empty @JsonTypeName() correctly.

Yona-Appletree:
Adds logic to handle empty value of @JsonTypeName as it's javadoc specifies


vojtechhabarta:
Thank you for this fix.
I just improved test and now [line 63](https://github.com/vojtechhabarta/typescript-generator/commit/230f7491f4a438d86f4ffb3a5007a0708cd260ef#diff-e74182671fc3cd59de8018a3618dbeffR63) actually tests this case.
Thanks again.

Issue #174, opened 2017-09-16
Jackson2: Support BeanPropertyWriter's metadata for optional properties


Yona-Appletree:
Currently, the Jackson2Parser only looks for optionalAnnoations when determining if a property should be optional. It seems like it should default to `! beanPropertyWriter.getMetadata().isRequired()` so as to support custom optional-detection by jackson modules.

In my case, I want it so that the Kotlin module can determine nullability.

vojtechhabarta:
Resolved in #183. Released [v1.29.355](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v1.29.355).

PR #179, opened 2017-09-18
Fix for #174 - Flag to use Jackson2's required for optional

Yona-Appletree:
Adds a flag which enables use of the Jackson2 optional concept to control the default optionality status for properties. Allows use of custom Jackson2 module logic to control optionality instead of relying on optional annotations alone. 

My use case is to allow correct handling of Kotlin's optional properties, for which there are no runtime-readable annotations generated.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -70,6 +70,7 @@ public class Settings {
     public String typescriptVersion = "^2.4";
     public boolean displaySerializerWarning = true;
     public boolean disableJackson2ModuleDiscovery = false;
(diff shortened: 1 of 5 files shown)


Yona-Appletree:
So, in the case of my project, this actually does work, I think because I don't have the Jaxb module or similar, I haven't looked too far into it. But when using the kotlin module, the required flag is set correctly, so will likely work for other Kotlin folks.

As you said, perhaps there could be a setting to specify which Jackson2 modules are included, by class name pattern list or similar.

vojtechhabarta:
@Yona-Appletree I reimplemented this feature with different configuration - added `optionalProperties`, and deprecated `declarePropertiesAsOptional`. I also changed how typescript-generator loads Jackson2 modules (motivated by `JaxbAnnotationIntrospector.hasRequiredMarker()` bug).

Could you please review new PR, at least Kotlin part?

PR #183, opened 2017-09-28
Optional properties using Jackson2

vojtechhabarta:
This PR is continuation of #179, thanks @Yona-Appletree for feature request and first implementation.

This PR adds `optionalProperties` parameter with allowed values `useSpecifiedAnnotations`, `useLibraryDefinition` and `all`.

It also changes how Jackson2 modules are treated, this is incompatible change, please see details below.

### 1. `optionalProperties` = `useSpecifiedAnnotations`
Let's say we want to generate following TypeScript interface:
``` typescript
interface Person {
    name: string;  // required
    age?: number;  // optional
}
```
We can do this with this configuration:
``` xml
<configuration>
    <optionalProperties>useSpecifiedAnnotations</optionalProperties>
    <optionalAnnotations>
        <annotation>javax.annotation.Nullable</annotation>
    </optionalAnnotations>
</configuration>
```
and Java class:
``` java
public class Person {
    public String name;
    @Nullable
    public int age;
}
```

This was already possible and it is still **default** behavior.

### 2. `optionalProperties` = `useLibraryDefinition`
With this new feature we can leverage Jackson2 to determine which properties are optional and which required. Then it is possible to use Jackson2 annotations, JAXB annotations or even Kotlin nullable types to specify property optionality.

#### Jackson2 library
Following configuration:
``` xml
<configuration>
    <optionalProperties>useLibraryDefinition</optionalProperties>
</configuration>
```
allows us to use `@JsonProperty` to specify optionality:
``` java
public class Person {
    @JsonProperty(required = true)
    public String name;
    @JsonProperty(required = false)
    public int age;
}
```

#### JAXB library
Similarly for JAXB. Configuration:
``` xml
<configuration>
    <jsonLibrary>jaxb</jsonLibrary>
    <optionalProperties>useLibraryDefinition</optionalProperties>
</configuration>
```
class:
``` java
public class Person {
    @XmlElement(required = true)
    public String name;
    @XmlElement(required = false)
    public int age;
}
```

#### Kotlin nullable types
Jackson2 provides `KotlinModule` which makes nullable types accessible to typescript-generator.
This module can be configured using this snippet:
``` xml
<configuration>
    <optionalProperties>useLibraryDefinition</optionalProperties>
    <jackson2Modules>
        <module>com.fasterxml.jackson.module.kotlin.KotlinModule</module>
    </jackson2Modules>
</configuration>
```
Then we can write Kotlin data class like this:
``` kotlin
data class Person(
        val name: String,
        val age: Int? = null
)
```
and typescript-generator will know from Jackson2 which properties are required/optional.

### 3. `optionalProperties` = `all`
This makes all properties optional. This setting have same effect as `declarePropertiesAsOptional` parameter (which is now deprecated).

## Jackson2 modules
Before this PR typescript-generator by default loaded all Jackson2 modules automatically. This automatic module discovery relies on module presence on classpath which is relatively fragile mechanism and for example `JaxbAnnotationModule` currently breaks `@JsonProperty.required` functionality.

So now typescript-generator by default loads no modules. Previous behavior can be turned on using `jackson2ModuleDiscovery` parameter. Individual modules can be specified using `jackson2Modules` parameter (recommended way). Old `disableJackson2ModuleDiscovery` parameter is effectively removed.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/OptionalProperties.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/OptionalProperties.java
@@ -1,0 +1,7 @@
+
+package cz.habarta.typescript.generator;
+
+
+public enum OptionalProperties {
+    useSpecifiedAnnotations, useLibraryDefinition, all
+}
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -1,6 +1,7 @@
 
 package cz.habarta.typescript.generator;
 
+import com.fasterxml.jackson.databind.Module;
 import cz.habarta.typescript.generator.emitter.Emitter;
 import cz.habarta.typescript.generator.emitter.EmitterExtension;
 import cz.habarta.typescript.generator.emitter.EmitterExtensionFeatures;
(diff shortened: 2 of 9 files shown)


Yona-Appletree:
This looks great, thanks. I've tested it on my Kotlin project and everything seems to work as expected.

Issue #191, opened 2017-10-27
Add support for API compatible Class based Enums


jtoplak:
Proper serialization/deserialization of enums is a huge waste of time.  As such, robust API safe enums are quite often mimicked using classes (and they should be in most cases). It avoids an incredible amount of hassle and is commonly implemented with the following pattern.

    @EqualsAndHashCode
    public class ApiFriendlyType {

        public static final ApiFriendlyType FOO = new ApiFriendlyType("FOO");
        public static final ApiFriendlyType BAR  = new ApiFriendlyType("BAR");

        private final String value;

        @JsonCreator public ApiFriendlyType(String value) { 
            this.value = Objects.requireNonNull(value);
        }

        @JsonValue public String getValue() { return value; }

        @Override public String toString() { return value; }

    }

Since the class is not treated like an enum, this pattern will always produce

    interface ApiFriendlyType {
        value: string;
    }

Instead of

    type ApiFriendlyType = "FOO" | "BAR";
    interface ApiFriendlyEnum {
        value: ApiFriendlyType
    }

or the preferred

    enum ApiFriendlyType {
        FOO = "FOO",
        BAR = "BAR"
    }

I decided to try and implement a custom TypeProcessor that checks for class names with the suffix "Type". I ran into issues when I tried to pass FOO and BAR and the TsType enum type to the Result object. As it is I also don't see any way to utilize an EmitterExtension.

How else could I properly map this kind of Pseudo Enum to a typescript enum?

vojtechhabarta:
@jtoplak this cannot be done using `TypeProcessor` or `EmitterExtension`. `TypeProcessor` maps one Java type to TypeScript type, it cannot produce new declarations. `EmitterExtension` can add declaration to output but cannot change declarations.

This could be achieved in `ModelParser` or `ModelCompiler` (https://github.com/vojtechhabarta/typescript-generator#architecture) but they are not extensible now. Would you be ok with using your own typescript-generator fork?

Do you think this feature is mostly for Jackson2 classes?

If yes you could extend `Jackson2Parser`, override `parseClass` and produce `EnumModel` insteadof `BeanModel` for your "enum" classes.

If you wanted this to work with all libraries you could create a method which transforms `TsModel`. In `ModelCompiler.javaToTypeScript()` is chain of such transformers. You could put it before enum section.

jtoplak:
@vojtechhabarta That's very helpful, thanks for the quick response. 

[...]

PR #192, opened 2017-11-08
Additive change for class based enum

jtoplak:
This feature is only applied to classes that have a naming convention that is unique to enums. This way, one does not have to conform on how class enums should be defined. The transformation takes place in the `ModelCompiler` and will work for both jackson2 and jaxb.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <groupId>cz.habarta.typescript-generator</groupId>
     <artifactId>typescript-generator</artifactId>
     <packaging>pom</packaging>
-    <version>1.29-SNAPSHOT</version>
+    <version>1.30-SNAPSHOT</version>
     <name>typescript-generator</name>
     <description>Generates TypeScript declaration file from specified java classes.</description>
     <url>https://github.com/vojtechhabarta/typescript-generator</url>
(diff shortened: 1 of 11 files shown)


vojtechhabarta:
@jtoplak, thanks for your PR.
I have some notes:
- In #191 you said that enum serialization and deserialization is slow. Could you add some test `main` method which would compare normal enums with class based enums? I guess the difference will be few percent if any 🙂 I briefly looked into Jackson2 sources and I think they have some cache for enums.
- I really like how you implemented this feature in `ModelCompiler`.
- Typescript-generator still supports Java 7. It is built using JDK 7 to ensure this compatibility. That's why the build failed.
- Do not add new compile dependencies.
- You don't have to change version, I will change it before releasing to maven central. If you changed it, it should be `1.30-SNAPSHOT`, not `1.30`. Since your PR is created from your `master` branch I think your last commit (c79102f2f96347cd054f2389e282c3c32dc1fe35) which changed the version wasn't meant for this PR. (It is better to create PRs in different branches).

jtoplak:
Thanks for the notes!
Sorry for taking so long, I didn't have the time to come back to this. 

Just to clarify, I didn't actually say that class based enums serialization was faster. I was more referring to the fact that serializing enums in a published API is a waste of time because it breaks with versioning. I just did a really bad job at explaining my argument :)  [Here](http://brian.pontarelli.com/2006/01/17/versioning-and-serialization-of-enums/) is an article that explains the problem better than I can. I cherry picked my commits to a branch and sent you a PR from a different branch. I will close this PR.

PR #194, opened 2017-11-12
Class based enum

jtoplak:
In some select situations enums are better treated as classes. For example, when developing micro-services and deploying on a regular basis it is common to add properties to existing enums which are published as DTO's for the micro-service API.  In this case the published client will be using an enum with a defined or 'finite' set of properties. As such, if a newer version of the micro-service with new enum properties was ever deployed whilst some client was still using the old version of the enum, the client will break. This can be avoided by using class based enums. They are API safe because the number of properties are not defined. If client was using an older version of the enum with more/less properties, the application would not break at runtime. [Here](http://brian.pontarelli.com/2006/01/17/versioning-and-serialization-of-enums/) is an article that does a pretty good job at explaining the problem.

This feature works for class enums that use `public static final` fields since this is best practice. 
An example of a class based java enum would be 

    public class DummyClassEnum {

        public static final DummyClassEnum ATYPE  = new DummyClassEnum("ATYPE");
        public static final DummyClassEnum BTYPE  = new DummyClassEnum("BTYPE");
        public static final DummyClassEnum CTYPE  = new DummyClassEnum("CTYPE");

        private final String value;

        @JsonCreator
        public DummyClassEnum(String value) { this.value = Objects.requireNonNull(value); }

        @JsonValue
        public String getValue() { return value; }

        @Override
        public String toString() { return value; }

    }

This feature is only applied to classes that have a chosen naming convention that is unique to class enums. This way, one does not have to conform on how class enums should be defined. The transformation takes place in the `ModelCompiler` and will work for both `jackson2` and `jaxb`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <groupId>cz.habarta.typescript-generator</groupId>
     <artifactId>typescript-generator</artifactId>
     <packaging>pom</packaging>
-    <version>1.29-SNAPSHOT</version>
+    <version>1.30-SNAPSHOT</version>
     <name>typescript-generator</name>
     <description>Generates TypeScript declaration file from specified java classes.</description>
     <url>https://github.com/vojtechhabarta/typescript-generator</url>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <groupId>cz.habarta.typescript-generator</groupId>
     <artifactId>typescript-generator</artifactId>
     <packaging>pom</packaging>
-    <version>1.30</version>
+    <version>1.29-SNAPSHOT</version>
     <name>typescript-generator</name>
     <description>Generates TypeScript declaration file from specified java classes.</description>
     <url>https://github.com/vojtechhabarta/typescript-generator</url>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <groupId>cz.habarta.typescript-generator</groupId>
     <artifactId>typescript-generator</artifactId>
     <packaging>pom</packaging>
-    <version>1.30-SNAPSHOT</version>
+    <version>1.30</version>
     <name>typescript-generator</name>
     <description>Generates TypeScript declaration file from specified java classes.</description>
     <url>https://github.com/vojtechhabarta/typescript-generator</url>
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>cz.habarta.typescript-generator</groupId>
         <artifactId>typescript-generator</artifactId>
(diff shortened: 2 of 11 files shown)


vojtechhabarta:
Thanks @jtoplak, now the motivation for this feature is clear. In some cases I would use `DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_AS_NULL` in Jackson but there are definitely cases where you want to pass (possibly) unknown values further which is not possible with normal enums.

As we discussed earlier (https://github.com/vojtechhabarta/typescript-generator/issues/191#issuecomment-339891761), implementing this feature as an extension wasn't possible. But I thought typescript-generator should have enough extensibility to support this so I created extension mechanism for `ModelCompiler`. It is now possible to extend abstract class `Extension` and provide `ModelTransformer` which can do `TsModel` transformation similarly as you implemented in this PR. I also added possibility to pass some configuration properties to these extensions. See `GenerateMojo.extensionsWithConfiguration` and `Extension.setConfiguration()`. Configuration in Maven is little bit verbose:
``` xml
<configuration>
    <extensionsWithConfiguration>
        <extension>
            <className>cz.habarta.typescript.generator.ext.SomeExtension</className>
            <configuration>
                <someProperty>someValue</someProperty>
            </configuration>
        </extension>
    </extensionsWithConfiguration>
    ...
</configuration>
```
but on the other hand now anybody can implement such kind of features.

Would you be interested to reimplement this feature as an extension? If yes, here is a sample how you can start:
``` java
package cz.habarta.typescript.generator.ext;
public class ClassEnumExtension extends Extension {
    @Override
    public void setConfiguration(Map<String, String> configuration) throws RuntimeException {}
    @Override
    public List<TransformerDefinition> getTransformers() {
        return Arrays.asList(new TransformerDefinition(ModelCompiler.TransformationPhase.BeforeEnums, new ModelTransformer() {
            @Override
            public TsModel transformModel(SymbolTable symbolTable, TsModel model) {}
        }));
    }
}
```
I think this feature would be an ideal candidate for first `Extension` 🙂

vojtechhabarta:
@jtoplak I merged your PR but I refactored the functionality into an `ClassEnumExtension`.
It can be used in Maven like this:
``` xml
<configuration>
    <extensionsWithConfiguration>
        <extension>
            <className>cz.habarta.typescript.generator.ext.ClassEnumExtension</className>
            <configuration>
                <classEnumPattern>ClassEnum</classEnumPattern>
            </configuration>
        </extension>
    </extensionsWithConfiguration>
    ...
</configuration>
```

It is little bit verbose but if you don't need configuration it can be shortened:
``` xml
<configuration>
    <extensions>
        <extension>cz.habarta.typescript.generator.ext.ClassEnumExtension</extension>
    </extensions>
    ...
</configuration>
```

Thanks again for your PR and sorry for not merging it as you intended.

PR #198, opened 2017-11-30
Respect outputs for compileGroovy tasks

jechlin:
So that you can write classes in groovy for translation to .d.ts.

Prior to gradle 4.0, this worked fine, as compileGroovy and compileJava
shared the same output path. After gradle 4.0, each language has a
different output directory.


vojtechhabarta:
@jechlin thanks for your PR.

PR #200, opened 2017-12-06
Fixed ignoring npm name and version properties

szymonbultrowicz:
It looks like current version of GenerateTask.java has a wrong condition when setting npmName and npmVersion properties. I assume it should work like that (if generateNpmPackageJson is set to true):
1. npmName/npmVersion not set -> use default value
1. npmName/npmVersion set -> use given value

However, currently it works like that:
1. npmName/npmVersion not set -> error is being thrown with message that those properties are required
1. npmName/npmVersion set -> uses default values (ignores the given ones)

This PR fixes that issue.

### DIFF
--- a/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
+++ b/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
@@ -144,8 +144,8 @@ public class GenerateTask extends DefaultTask {
         settings.loadIncludePropertyAnnotations(classLoader, includePropertyAnnotations);
         settings.loadOptionalAnnotations(classLoader, optionalAnnotations);
         settings.generateNpmPackageJson = generateNpmPackageJson;
-        settings.npmName = npmName != null && generateNpmPackageJson ? getProject().getName() : npmName;
-        settings.npmVersion = npmVersion != null && generateNpmPackageJson ? settings.getDefaultNpmVersion() : npmVersion;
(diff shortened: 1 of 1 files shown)


vojtechhabarta:
You are right, it's wrong.

You are also right that it should generate default value when
1. the value is not filled and 
2. `generateNpmPackageJson` is `true`.

The second part is missing in your PR.

I think that what's needed is just to flip `null` tests. Then if `generateNpmPackageJson` is `false` but value is `null` then it stays `null`.

See this in Maven plugin here: https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-maven-plugin/src/main/java/cz/habarta/typescript/generator/maven/GenerateMojo.java#L578-L579

What do you think?

szymonbultrowicz:
Actually, I think there is no need to check if the flag is true. Without it, the condition is simpler and those values are going to be ignored, anyway. Although, if you think it's better to check it, I'll do it, no problem.

vojtechhabarta:
Currently if `generateNpmPackageJson` is `false` the values are NOT silently ignored but `Settings.validate()` complains that they can only be used when `generateNpmPackageJson` is `true`. I know it's rather pedantic 😄  but I wanted to give users detailed feedback when they're configuring typescript-genereator.

See https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java#L270

szymonbultrowicz:
Got it. Fixed.

vojtechhabarta:
Merged. Thank you very much.

PR #207, opened 2018-01-03
WIP: Generate optional properties as null

JamesGuthrie:
Currently, type interfaces which are generated from JAX-RS endpoints translate properties with a `@Nullable` annotation into an optional field as follows:

```
interface Foo {
  bar?: string
}
```

Whereas the actual object sent/expected by a JAX-RS endpoint would have the form:

```
interface: Foo {
  bar: string | null
}
```

This change allows the emitter to be switched from the first representation to the second by setting `optionalAsNull` to true.

Note:
- I'm hoping to start a discussion with this PR, it's not ready to be merged yet (hence WIP).
- I have't added tests for the new functionality because I'm not sure if this is the best way to implement it, or what the approach is with testing the impact of different settings. Also, I'm not too sure about other impacts on code/functionality that I'm not using.
- I'm not sure about the name of the parameter, maybe `optionalPropertiesAsNull` would be more explicit.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -34,6 +34,7 @@ public class Settings {
     private Predicate<String> excludeFilter = null;
     @Deprecated public boolean declarePropertiesAsOptional = false;
     public OptionalProperties optionalProperties; // default is OptionalProperties.useSpecifiedAnnotations
+    public boolean optionalAsNull = false;
     public boolean declarePropertiesAsReadOnly = false;
     public String removeTypeNamePrefix = null;
     public String removeTypeNameSuffix = null;
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsProperty.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsProperty.java
@@ -23,7 +23,7 @@ public class TsProperty {
     }
 
     public String format(Settings settings) {
-        final String questionMark = (tsType instanceof TsType.OptionalType) ? "?" : "";
+        final String questionMark = (tsType instanceof TsType.OptionalType && !settings.optionalAsNull) ? "?" : "";
         return Emitter.quoteIfNeeded(name, settings) + questionMark + ": " + tsType.format(settings) + ";";
     }
 
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
@@ -200,7 +200,11 @@ public abstract class TsType implements Emittable {
 
         @Override
         public String format(Settings settings) {
-            return type.format(settings);
+            if (settings.optionalAsNull) {
+                return type.format(settings) + " | null";
+            } else {
+                return type.format(settings);
+            }
         }
 
     }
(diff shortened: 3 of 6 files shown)


vojtechhabarta:
Thanks for this PR.

Using `| null` is probably right solution for optional properties when using `strictNullChecks` compiler switch and you are right that it should be configurable because Jackson/Jersey can be configured to skip properties with `null` values in which case we need current behavior.

But I was also thinking about nullability in general. In Java almost everything can have `null` value. So for example `List<String>` should be in fact mapped to something like `(string | null)[] | null` which is not much readable and it is also harder to use. And this becomes even more complicated for example with `List<Map<String, String>>` and so on. It would also complicate typescript-generator itself because it should also work with generics, primitive types should be handled specifically, property optionality is different from nullability nested in some arrays or objects and so on. So I decided not to do this (year ago).

I also think that current solution is not so bad because users can test values using `== null` or `== undefined` which works regardless `strictNullChecks` switch and regardless how Jackson handles Java `null` values (produces `null` in JSON or skips properties). In opposite direction I think Jackson also works with both cases (`null` or missing) so user can just always use `undefined`.

But maybe it's time to revisit this decision. I think we have these options:
1. implement nullability everywhere
2. just change optional properties as you suggested
3. leave it as it is now

If you don't mind I would wait for more feedback for some time before making any decision.

So any comments?

JamesGuthrie:
Thanks for the detailed response.

Exactly, my use-case here is that we are wanting to activate `strictNullChecks`, which doesn't work well with nullable fields as optional (`?`).

While I agree with you that Java allows almost everything to have a null value, my use-case is specific cases in which a property of the JAX-RS endpoint's return type is annotated with `@Nullable`. As Java doesn't allow the option to opt-out of nullability, I see expressly declaring a property as nullable to be more of an indication as to what may/may not be null (i.e. if I haven't declared a property to be `@Nullable`, then I don't expect it to be null).

I guess you're correct that it is possible to work around the current implementation with an `== null` or `== undefined` check, but that is only going around the problem that the generated interface does not accurately represent the type which is actually being passed from backend to frontend.

I'm not sure what the implications of 1. are, but I would prefer to not do 3. because it doesn't make anything better for me. Is 2. not possibly a good middle-ground between 1. and 3?

vojtechhabarta:
Adding nullability to just optional properties (middle ground) is probably good option.

Since Jackson accepts both 1. missing properties and 2. `null` values and in opposite direction it can also 1. skip properties or 2. write `null` values (this can also be set per class using annotations) I think it could be useful to allow declare properties as both optional (`?`) and nullable (`| null`) at the same time (in addition to the original variants).

Example:

``` typescript
interface Something {
    optionalProperty1?: string;
    optionalProperty2: string | null;
    optionalProperty3?: string | null;
}
```

Third variant can not only describe well contract but it can also be more convenient for users because marking optional properties with `?` allows to skip them when creating object literals in TypeScript. Otherwise users would receive an error for example like this:

``` typescript
const something: Something = {
    // Error: Property 'optionalProperty2' is missing in type ...
    optionalProperty3: null
}
```

So we could add for example parameter `optionalPropertiesDeclaration` with values:
- `optionalMark` (default)
- `nullableType`
- `optionalMarkAndNullableType`

Another point is if we should use union types (`| null`) directly or create type alias (`Nullable<T>`). What do you think?

Note about implementation: logic for optional properties should not be done in `Emitter`, it should be placed in `ModelCompiler` and use `TsType` classes like `OptionalType`, `UnionType`, maybe `NullableType` (I am not sure).

I can implement this if you want.

vojtechhabarta:
I implemented `optionalPropertiesDeclaration` parameter with possible values:
- `questionMark`
- `nullableType`
- `questionMarkAndNullableType`

PR #226, opened 2018-04-14
feat(expanded-enum-values): Expand @JsonValue scenarios for enums

mike-plummer:
Currently enums support generating by `name`, `@JsonProperty` annotation on constant, or `@JsonValue` annotation on read method. I've encountered scenarios where code utilizes `@JsonValue` on enum instance fields or non-accessor methods (like `toString`) and propose expanding custom enum value support to these scenarios.

This PR expands support to enums like the following:

```java
public enum ToStringValuedEnum {
    A, B, C;

    @JsonValue
    @Override
    public String toString() {
        return "_" + name();
    }
}  // -> "_A", "_B", "_C"

public enum StringPropertyValuedEnum {
    A, B, C;

    @JsonValue
    private final String value = "_" + name();
}  // -> "_A", "_B", "_C"

public enum NumberPropertyValuedEnum {
    A, B, C;

    @JsonValue
    private final Integer value = ordinal();
}  // -> 0, 1, 2
```

## Performance Impact
Logic changes will increase the number of methods being checked per-enum and also will evaluate more fields than before thus it will technically be slower. This impact should be negligible considering the number of instance fields and methods on enums are typically minimal, and I believe the increased capability more than offsets any slowdown.

## Compatibility
This will modify generated schemas for any baselines making use of the scenarios above. I believe any situations where a change occurs would be desired since the schema would be updating to more accurately reflect what is serialized via JSON. I'm also not aware of any new edge cases that might cause issues since `@JsonValue` was already supported in more limited conditions.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
@@ -17,7 +17,7 @@ import cz.habarta.typescript.generator.compiler.EnumMemberModel;
 import cz.habarta.typescript.generator.util.Predicate;
 import java.beans.BeanInfo;
 import java.beans.Introspector;
-import java.beans.PropertyDescriptor;
+import java.beans.MethodDescriptor;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AccessibleObject;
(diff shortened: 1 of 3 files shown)


vojtechhabarta:
Thanks for this PR.
I appreciate how isolated the change is so it is easy to review.
Thanks again.

PR #228, opened 2018-04-17
Use @ApiModelProperty attribute dataType to override target type

littleli:


vojtechhabarta:
Ahoj 😉
thanks for your PR.

Issue #167, opened 2017-09-06
PLEASE VOTE: How do you run typescript-generator (currently or would like)?


vojtechhabarta:
Vote using 👍

vojtechhabarta:
# Maven

vojtechhabarta:
# Gradle

vojtechhabarta:
# Direct Java call

vojtechhabarta:
# Ant

garysieling:
SBT

thepian:
Gradle. But need to output as a class.

wimdeblauwe:
Maven, but from the command line. I don't want to have it integrated in my build, just run manually now and then and use the generated classes as a start point.

I currently configure the plugin in my `pom.xml` in the `<pluginManagement/>` section and then run it with `mvn cz.habarta.typescript-generator:typescript-generator-maven-plugin:generate`

joshiste:
From npm scripts.

AleksandarTokarev:
@vojtechhabarta Maybe u can add @wimdeblauwe 's answer in the wiki along with `mvn cz.habarta.typescript-generator:typescript-generator-maven-plugin:generate`. Really helpful stuff

PR #242, opened 2018-07-09
Add ability to scan for classes with specific annotations

IdanE:
Adds the parameter `classesWithAnnotations` which will run a scan on the classpath and retrieve only the classes with the given annotations.

Example usage:
````xml
    <configuration>
        <classesWithAnnotations>com.example.annotation.Expose</classesWithAnnotations>
        ...
    </configuration>
````

````java
@Expose
public class A {
    ...
}
````


vojtechhabarta:
@IdanE thanks for your PR.

I have one request: could you please return original `Input.fromClassNamesAndJaxrsApplication` method overload for backward compatibility, create new overload and delegate one to the other?

In the past I only cared about Maven and Gradle plugins compatibility but now I feel we should try to keep Java API reasonably compatible too because people also call typescript-generator directly, see #167.

IdanE:
Thanks for the feedback, amended.

Is that more or less what you meant?

vojtechhabarta:
Exactly, that's it.
I just changed the description to be more consistent with other similar parameters.
Thanks.

Issue #236, opened 2018-06-13
Convert to ES6 Map


fridoo:
I would like to convert a Java Map to an ES6 Map. 

From Java:
`public Map<String, Object> someMap;`

To Typescript:
`someMap: Map<string, any>;`

instead of
`someMap: { [index: string]: any };`

Is this possible somehow?

vojtechhabarta:
There is `customTypeMappings` parameter but it doesn't work with generics.
I'm thinking about improving it.

vojtechhabarta:
Finally I returned to this issue and it is now possible to use generics in `customTypeMappings` parameter.
Released in [v2.10.466](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.10.466).

vojtechhabarta:
Unfortunately mentioned change in [v2.10.466](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.10.466) broke mapping of generic classes to TypeScript non-generic types like `string`. So in [2.16.538](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.16.538) release I changed `customTypeMappings` parameter format (#384) and it is now needed to specify generic classes including their generic parameters.

PR #246, opened 2018-07-19
Transform Java Maps to ES6 Maps

jdussouillez:
References #236

The Java Map objects are now transformed into Map and not into indexed array.


vojtechhabarta:
@jdussouillez we cannot change Java Map transformation to ES6 Maps.

When JSON is parsed using `JSON.parse` function then maps are loaded into JavaScript objects so corresponding TypeScript types are indexed types (for example `{ [index: string]: string }`). This doesn't look as nice as ES6 Maps but it describes reality.

jdussouillez:
Oh sorry, I forgot the fact that Map cannot be handled easily in JSON.
But thank you for the quick answer !

Issue #248, opened 2018-07-27
Conditionally use non-const enums


danielkaneider:
As far as I can see there is an option to make all the enums const or not. In terms of performance we are using the const version for all. It turns out that there are some situations were some of the enums should preferably non-const. If we agree on the API I'm willing to provide a PR for that. 
My proposal:
- add a `nonConstEnumAnnotations : List` setting which forces enums to be exported as non-const. Since const enums seem to be the default one, an opposite annotation like (constEnumAnnotations) is not necessarily needed

vojtechhabarta:
Yes, since "const" is default, I would add `nonConstEnumAnnotations` next to `nonConstEnums` parameter in `Settings`, `GenerateMojo` and `GenerateTask`.

And I would add something like `isNonConstEnum` to `TsEnumModel` and move the logic from `Emitter` to `ModelCompiler`.

PR #249, opened 2018-07-30
Conditionally use non-const enums based on annotation (#248)

danielkaneider:
See discussion in issue #248

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -14,6 +14,8 @@ import java.net.URLClassLoader;
 import java.util.*;
 import java.util.regex.Pattern;
 
+import javax.management.ImmutableDescriptor;
+
 
 /**
  * See cz.habarta.typescript.generator.maven.GenerateMojo
@@ -48,6 +50,7 @@ public class Settings {
     public DateMapping mapDate; // default is DateMapping.asDate
     public EnumMapping mapEnum; // default is EnumMapping.asUnion
     public boolean nonConstEnums = false;
+    public List<Class<? extends Annotation>> nonConstEnumAnnotations = new ArrayList<>();
     public ClassMapping mapClasses; // default is ClassMapping.asInterfaces
     public List<String> mapClassesAsClassesPatterns;
     private Predicate<String> mapClassesAsClassesFilter = null;
@@ -138,6 +141,13 @@ public class Settings {
                 this.extensions.add(emitterExtension);
             }
         }
+    }
+
+    public static List<Class<? extends Annotation>> loadAnnotations(ClassLoader classLoader, List<String> stringAnnotations) {
+        if (stringAnnotations != null) {
+            return new ArrayList<>(loadClasses(classLoader, stringAnnotations, Annotation.class));
+        }
+        return new ArrayList<>(0);
     }
 
     public void loadIncludePropertyAnnotations(ClassLoader classLoader, List<String> includePropertyAnnotations) {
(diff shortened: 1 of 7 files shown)


typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
```diff
@@ -14,6 +14,8 @@
 import java.util.*;
 import java.util.regex.Pattern;
 
+import javax.management.ImmutableDescriptor;
```
vojtechhabarta: this import is unused

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
```diff
@@ -140,6 +143,13 @@ public void loadExtensions(ClassLoader classLoader, List<String> extensions, Lis
         }
     }
 
+    public static List<Class<? extends Annotation>> loadAnnotations(ClassLoader classLoader, List<String> stringAnnotations) {
```
vojtechhabarta: could you please replace this method with `loadNonConstEnumAnnotations` which would be consistent with other `load*` methods?

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/emitter/TsEnumModel.java
```diff
@@ -12,6 +12,7 @@
 
     private final EnumKind kind;
     private final List<EnumMemberModel> members;
+    private boolean isConstEnum = true;
```
vojtechhabarta: could you please replace this field with `isNonConstEnum` field?
it doesn't look as good as your field but since "const" enums are default I would like to use consistently "nonConst" on all places

second thing is that this field should be final (these classes are meant to be immutable)

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
danielkaneider: This one would actually now be more generic, since it returns those annotations instead of setting them on a specific field. This could replace some of the others actually

vojtechhabarta: ok, leave your method as it is now and I will use it on other places too, or maybe I will change `loadClasses` to return empty list instead of `null` and use it directly

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/emitter/TsEnumModel.java
```diff
@@ -12,7 +12,7 @@
 
     private final EnumKind kind;
     private final List<EnumMemberModel> members;
-    private boolean isConstEnum = true;
+    private boolean isNonConstEnum = false;
```
vojtechhabarta: could you please mark this field `final` and make this class immutable again?

danielkaneider: About the final thing, it would break backwards-compatibility, since I cannot use getters/setters any more, but I'll need to pass the field to the constructor.

vojtechhabarta: I think we can have two constructors: one for backward compatibility and one new with all the fields.

And instead of setter we can have `withIsNonConstEnum(boolean)` or just `asNonConst()` that would return new instance.

vojtechhabarta:
I think we are done.
Should I squash and merge it?

danielkaneider:
Sure, thx

Issue #250, opened 2018-08-02
Primitive data types should not be optionals


danielkaneider:
Hi, 
we're currently using `OptionalProperties.all` which isn't ideal for all cases. To be more specific:
- primitives such as `int getNumber()` should not be optional
- all other objects eg `Integer getNumber()` should be by default optional
- eventually a _NonNull_ annotation should be used to mark some returned Object as non-optional

If you want I can make a PR for that. I'd add an additional annotation `OptionalProperties.default`, which could easily cover the first two cases. For the last case we could either use the `optionalAnnotations` setting for the non-optional annotations, or create a new one `nonOptionalAnnotations`.

vojtechhabarta:
Hi,

you can also use `OptionalProperties.useLibraryDefinition` which determines properties "optionality" from configured library (I guess you are using Jackson2). By default Jackson2 considers properties as optional unless they are annotated with `@JsonProperty(required = true)`.

I think this is close to your needs. It satisfies point (2) - properties are optional by default, for point (3) you can use `@JsonProperty` annotation, for point (1) you also need to use `@JsonProperty` annotation in this case.

So only point (1) doesn't work automatically as you wanted but on the other hand it is consistent with Jackson2 library which also considers properties of primitive types as optional by default.

#183 contains more information about this parameter.

What do you think?

danielkaneider:
You're mostly right. Jackson2 is a good starting point. However Jackson(2) can be configured globally, while the typescript-generator only considers the annotations of the current class. Jackson has a `Include` annotation which can be set among others to `NON_NULL` or `NON_EMPTY`. In the first case a int type would be non-nullable, while in the second case it is not. Sure, you could set the annotations on every primitive method, but generally you don't do that. Therefore it might be good to have some option in the generator for primitives.

danielkaneider:
Nevermind, I found a working solution. Using a extension using  `BeforeEnums` I could do a workaround for those primitives. During implementing that I found, that it would be nice to have an extension point before the `processModel` method is called (`TransformationPhase.BeforeModels`) I.e. allowing an extension to make modifications to the model, before they are converted to the internal TS model. Besides that, feel free to close this issue.

vojtechhabarta:
Your PR is now merged and released, closing the issue.

PR #255, opened 2018-08-21
Add extension point before TsModel generation

danielkaneider:
As a finding of #250 I extended the ExtensionTransformer to be able to transform not only a `TsModel` but also a `Model`. This for example now allows to easily add class or method comments to the model, or perform some changes to the original model before it is actually converted to TS.


danielkaneider:
Done

Issue #142, opened 2017-05-12
Classloader issues cause maven plugin (v1.14.251) to fail when using classesFromAutomaticJaxrsApplication


sayresVT:
Using 1.14.251 of this plugin, we've been running into sporadic maven build issues that appear to be related to the classloading mechanism that's used when the classesFromAutomaticJaxrsApplication flag is set to `true`.  When running our multi-module maven build in parallel, the plugin will occasionally attempt to load a `@Path` annotated class that is from another module that isn't a direct dependency of the project running the plugin.  When it attempts to load the class, a `java.lang.ClassNotFoundException` is thrown and the build fails.

We had attempted to add the package of the class to the excludeClassPatterns, but that didn't help due to the fact that the Input/JaxrsApplicationScanner classes perform their classloading before any filtering occurs.  This is problematic because the FastClasspathScanner is occasionally finding classes that aren't available to the thread context class loader.  I was thinking there might be a couple of ways to address this:

1.  Perform class exclusion filtering on the FastClasspathScanner using the [constructor scan spec](https://github.com/lukehutch/fast-classpath-scanner/wiki/2.-Constructor#scan-spec).
2.  Perform exclusion filtering on class names found by the FastClasspathScanner before attempting to load them
3.  Allow individual classloading failures to be suppressed without causing the entire plugin execution to fail

We can update to the latest version of the plugin, but from what I can tell, the classpath scanning/classloading mechanism hasn't changed in a way that would prevent us from encountering this issue.  This is the relevant stack trace from our CI server when this issue causes the build to fail:
```
12:20:30 org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal cz.habarta.typescript-generator:typescript-generator-maven-plugin:1.14.251:generate (generate) on project PROJECT: Execution generate of goal cz.habarta.typescript-generator:typescript-generator-maven-plugin:1.14.251:generate failed: Cannot load JAX-RS application. For more information see https://github.com/vojtechhabarta/typescript-generator/wiki/JAX-RS-Application.
12:20:30 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
12:20:30 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
12:20:30 	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
[...]

PR #261, opened 2018-09-01
port to latest ClassGraph library

cjnygard:
- update code to use ClassGraph scanner (FastClasspathScanner is deprecated)

### DIFF
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -58,9 +58,9 @@
             <version>2.1</version>
         </dependency>
         <dependency>
-            <groupId>io.github.lukehutch</groupId>
-            <artifactId>fast-classpath-scanner</artifactId>
-            <version>2.9.4</version>
+            <groupId>io.github.classgraph</groupId>
+            <artifactId>classgraph</artifactId>
+            <version>4.1.5</version>
         </dependency>
         <!--test dependencies-->
         <dependency>
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Input.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Input.java
@@ -4,12 +4,15 @@ package cz.habarta.typescript.generator;
 import cz.habarta.typescript.generator.parser.*;
 import cz.habarta.typescript.generator.util.Predicate;
 import cz.habarta.typescript.generator.util.Utils;
-import io.github.lukehutch.fastclasspathscanner.FastClasspathScanner;
-import io.github.lukehutch.fastclasspathscanner.scanner.ScanResult;
+import io.github.classgraph.ClassGraph;
+import io.github.classgraph.ClassInfoList;
+import io.github.classgraph.ScanResult;
+
 import java.lang.reflect.*;
 import java.net.URLClassLoader;
 import java.util.*;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 
 public class Input {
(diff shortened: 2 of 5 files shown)


vojtechhabarta:
@cjnygard thanks for your PR which solves issue #257.

I have a comment regarding performance. Class-path scanning is relatively expensive operation so I wanted to keep its impact minimum and only use it when it is really needed. Typescript-generator has several ways how to specify input classes. Following table summarizes which way uses class-path scanning.

| parameter | class-path scanning required |
| --- | --- |
| classes | No |
| classPatterns | Yes |
| classesWithAnnotations | Yes |
| classesFromJaxrsApplication | No |
| classesFromAutomaticJaxrsApplication | Yes |

So class-path scanning should be invoked only when some of the marked parameters is used and it should be invoked only once if more of these parameters are used. That was the purpose of `Input.ClasspathScanner` class which implemented this requirement by lazy invoking the scanning.

I would also like to keep the log message with statistics how long scanning took and how many classes are on class-path.

There was also used `overrideClasspath` which I added when we were solving problem with parallel build (#142) along with possibility to turn on verbose output.

cjnygard:
I understand what you are asking for.  However, some differences in how ClassGraph is implemented may prevent what you're trying to achieve.  Perhaps this is my flawed interpretation, but here's a quick summary:

- ClassGraph makes specific recommendations not to mix how you load the class info.  In particular, you must not mix `Class.forName()` with `ClassInfo.loadClass()` or you'll get inconsistent results and exceptions (https://github.com/classgraph/classgraph/wiki/Code-examples#get-class-references-for-subclasses-of-comxyzcontrol)
- ClassGraph docs recommend doing the scan inside a try-with-resources so it's guaranteed to be cleaned up properly.  
- ClassGraph builder pattern only allows `verbose()` without a flag.

So the first point implies we need to load all classes through ClassGraph if we need to load any through ClassPath.  I gave some thought to trying to avoid ClassGraph if possible, but it seemed like some crufty code to look for specific combinations of options and (even worse) go through separate but functionally identical code paths to load the configured classes or JaxrsApplication classes two different ways.

Also, the third point about `verbose()` also makes setting that within a try-with-resources difficult to do cleanly as an optional parameter.  I ran it once with verbose on, and the output was... comprehensive (copious, verging on TMI).

Unless I'm missing something, the try-with-resources should take care of loading the ClassGraph just once, without the need for a separate class, in a safe way.  Unfortunately, that construct makes some of your other requests difficult.

I'm open to recommendations.  In the meantime, I added the classpath override and the timing output.  One thing I'm curious about, the code passed unit tests without the classpath override, so I thought it might be unnecessary since the code already resets the classloader via `Thread.currentThread().setContextClassLoader(classLoader)` just before the ClassGraph scan.

BTW, on my macbook, it finds 27K+ classes in 1.4 secs.  Not sure if that's considered good enough or not... but if you don't need all info (members, types, etc.) some info can be skipped and the scan time reduced.

vojtechhabarta:
**Loading classes:** yes, mixing class loaders is just wrong. Typescript-generator loads classes using class loader specifically created by particular build plugin and it uses `FastClasspathScanner` only for getting list of class **names** so there should not be any mix of class loaders. I would like to keep it this way because it works well and I don't want to break people's builds. There should be better test coverage but it is hard to write unit tests for this kind of functionality. However current implementation has been tested really well by real usage. I also remember how hard was to solve a problem with class loading in parallel build year ago. So for these reasons I really prefer changes to be minimum (assuming `ClassGraph` works the same way as `FastClasspathScanner`).

**Try-with-resources:** it is just a shorter syntax for try-finally so you have some options how to deal with it. You can replace it with your own call to `close` method in `finally` block. Or you can write method which builds the resource and call this method from `try` resource declaration like this:
``` java
try (ScanResult scanResult = scanClasses(debug)) {
    // ...
}

private static ScanResult scanClasses(boolean verbose) {
    ClassGraph classGraph = new ClassGraph();
    classGraph = classGraph.enableAllInfo();
    if (verbose) {
        classGraph = classGraph.verbose();
    }
    final ScanResult scanResult = classGraph.scan();
    return scanResult;
}
```

**Verbose:** previous example also shows how to conditionally set verbose flag when there is no method with `boolean` parameter for it.

**Performance:** according to my test as well as your numbers it seems that `ClassGraph` works similarly to `FastClasspathScanner` so that's good. But if we can make it work significantly faster by changing one line (`enableAllInfo`) I would do it.

cjnygard:
fixed as noted.

Issue #262, opened 2018-09-02
How can I use field name instead of getter/setter based name?


chinhodado:
I have a class

```
class A {
    boolean isGood;
    public boolean isGood() {return isGood;}
    public void setGood(boolean good) {this.isGood = good}
}
```

The generated TypeScript is

```
export interface A {
    good?: boolean;
}
```

How can I make it so that the field name `isGood` is used instead?

vojtechhabarta:
If you are using Jackson you can use `@JsonProperty` annotation. There are also annotations like `@JsonAutoDetect`, `@JsonIgnore`.

If you don't know how to use these annotations I can help you more.

chinhodado:
Thank you, I know about those annotations but the issues is 

 - I have hundreds of classes so going through all of them to add annotations is annoying. 
 - Also, I use GSON as the main JSON library so adding annotations from another library just for this purpose is rather not nice. It clutters the code.

I have opened a PR for this: https://github.com/vojtechhabarta/typescript-generator/pull/263

It's a simple setting and it works well.

vojtechhabarta:
I think it would be good if typescript-generator supported GSON directly.
While your PR is simple trying to make Jackson work the same way as GSON is hacky and could soon become complicated. I don't know much about GSON, maybe it is really subset of Jackson but I would still prefer to have first-class support for it.

chinhodado:
I'm not exactly trying to make it "work the same way as GSON". Using fields instead of getter/setters is a normal config in most JSON libraries, it's not specific to GSON or anything. You can see [here](https://stackoverflow.com/questions/7105745/how-to-specify-jackson-to-only-use-fields-preferably-globally) that it's a very common question for jackson. 

Even with GSON support or not, I would still want to have this option for jackson. Alternatively, you can change the API to optionally take in a custom ObjectMapper object and everyone can set whatever config they want.

vojtechhabarta:
Yes, I agree that being able to specify Jackson global configuration would be useful feature. So I was also thinking about taking `ObjectMapper` as a parameter but that would only work when invoked from Java. I would like to have something that would also work in Maven and Gradle plugins without adding many top-level parameters. Maybe we could have some `jackson2Configuration` parameter which would be object that could have these settings and could grow in the future.

vojtechhabarta:
I will implement it.

chinhodado:
Awesome, thank you!

vojtechhabarta:
[...]

PR #263, opened 2018-09-03
Add settings for only using fields when using Jackson

chinhodado:


vojtechhabarta:
Implemented slightly differently, see comments in #262.

PR #270, opened 2018-09-22
Class enum compatibility with regular enums

jtoplak:
I haven't seen this codebase since it's been merged with the [class enum PR](https://github.com/vojtechhabarta/typescript-generator/pull/194) as an extension. Very nice! 

I have recently started using the `ClassEnumExtension` in another project and I found an edge case where there are both enums and class based enums that contain the `classEnumPattern`. In this case, the regular enums do not get processed at all. Now it also returns enum models in the `TransformerDefinition`.

Also, I fixed the .gitignore for `.iml` files since it was getting picked up on my local.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,7 @@ nbactions.xml
 
 # JetBrains IntelliJ
 .idea/
-**.iml
+*.iml
 
 # VS Code
 .vscode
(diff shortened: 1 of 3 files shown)


vojtechhabarta:
@jtoplak thanks for your PR!

PR #272, opened 2018-09-23
expose noTsLintDisable setting to plugins

jtoplak:
Prior to release `v2.1.410`, the generated file would be linted which in some cases is preferred. This setting was not being exposed through the plugins and I can no longer lint my generated file if I choose.


jtoplak:
@vojtechhabarta if and when this is merged do you think you could cut a release? It would be nice to use this soon. Not linting my ts hurts my ocd :)

vojtechhabarta:
To be honest I don't like adding yet another parameter to already long list.

Before `tslint:disable` comment was generated people sometimes complained that TSLint reports rule violations. But I think typescript-generator goal is not to be able to produce all different styles people might be using. I really think generated code (as well as npm dependencies) should not be linted.

But since you really want this feature I will merge it :-)

BTW if you would find solution in your build (like removing first line in generated file) before next version is released we could revert this change ;-)

PR #274, opened 2018-09-24
@JsonValue doesn't correctly work for number-typed methods

danielkaneider:
`@JsonValue` can be indicated to set a custom value for enums. It works fine for strings. But it appeared, that for number it only worked on a field, but not on a method.
This PR one adds a fix + some tests for that (the first test, already worked without the fix)


vojtechhabarta:
@danielkaneider thanks for your PR!

PR #291, opened 2018-11-09
Ignore BeanParam parameters when looking for EntityParameter

richturner:
When trying to determine the entity parameter for JAX-RS methods the BeanParam annotation should also be ignored. I have not compiled this PR but it is a simple change from quickly looking at the source code.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/JaxrsApplicationParser.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/JaxrsApplicationParser.java
@@ -29,7 +29,7 @@ import javax.ws.rs.core.GenericEntity;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.StreamingOutput;
-
+import javax.ws.rs.BeanParam;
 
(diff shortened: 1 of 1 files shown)


vojtechhabarta:
Thanks for this PR. Merged.

richturner:
Thanks for merging this so quick! any time scale on a release with this in?
Would be nice to keep standard maven dependency in my build script.

Kind Regards,

Rich

vojtechhabarta:
This was released in [v2.9.456](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.9.456) I just somehow forgot about this PR when releasing.

krezovic:
This breaks some things for us. You can annotate a JAX-RS method parameter with @BeanParam and it can be of any class type which can hold a lot of query parameters inside it, all of them annotated with @QueryParam. We relied on this design to properly type query params inside a single object. Any way to make it configurable?

```
class SearchParams {
  @QueryParam("id")
  private Integer id;

  @QueryParam("name")
  private String name;
}

@GET
public List<Object> getItems(@BeanParam SearchParams params) {
  return Collections.emptyList();
}
```

Now query params used in this way are never generated in the resulting TS file, nor the bean param class.

vojtechhabarta:
I created issue #318 for this case.

PR #297, opened 2018-12-03
Close class loader once finished gradle task

richturner:


vojtechhabarta:
This change is released in [v2.10.466](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.10.466).

Issue #285, opened 2018-11-01
Feature request: including classes that implement an interface


clamothe:
Allow a configuration in which all classes that implement a particular interface are included.

`          ScanResult.getClassesImplementing("com.example.ApiObject")
`

This is handy to use in-hand with polymorphic Jackson:

`
@JsonIgnoreProperties(ignoreUnknown = true)
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY)
public interface ApiObject {}
`

I'll create a PR for this in time if no one beats me to it.

vojtechhabarta:
Ok, so do you plan to add `classesImplementingInterfaces` next to `classesWithAnnotations` parameter?

clamothe:
Something like that. The implementation should probably support specifying
either a class or interface, of which all descendents will be included.

clamothe:
@antch That's a great suggestion. 

Could you elaborate on the `getSuperClass()` option?

clamothe:
@antch 

I wasn't able to get `classWithAnnotations` to work as described with an `@Inherited` annotation. I annotated an interface with it, but the implementations of the interface are not included. Perhaps this only works with a class?

vojtechhabarta:
Yes, I tried foillowing code and annotation inheritance only worked when `ApiObject` was abstract class (not interface):
``` java
System.out.println(ApiObject.class.getAnnotation(GenerateTypescript.class));
System.out.println(SomeApiObject.class.getAnnotation(GenerateTypescript.class));
```

I also found this in javadoc of `@Inherited` annotation:
> ... Note also that this meta-annotation only causes annotations to be inherited from superclasses; annotations on implemented interfaces have no effect.

clamothe:
Ah, interesting. That explains that.

vojtechhabarta:
@clamothe anything new? I am cleaning issues after release, can I close this issue or are you still interested in creating PR?

clamothe:
Thanks for the reminder. PR created

clamothe:
Thanks for merging. Would it be possible to have a release that includes this in the next week or two?

vojtechhabarta:
@clamothe I will release it till the end of this week.

clamothe:
Thank you!

vojtechhabarta:
Released in version [2.12.476](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.12.476).

PR #301, opened 2019-01-08
classesImplementingInterface

clamothe:
Support for new 'classesImplementingInterface' per request https://github.com/vojtechhabarta/typescript-generator/issues/285


vojtechhabarta:
Thanks for the PR.
I have one request: second commit is not needed since `ClasspathScanner.scanClasspath()` caches the result so classpath is scanned only when needed and only once. So could you please recreate the PR without this commit? Maybe we can rename `scanClasspath` method to for example `getScanResult` to make it more clear.

clamothe:
Done

vojtechhabarta:
Thanks. Merged.

PR #305, opened 2019-01-17
Preserve indentation, allowing code examples to be formatted correctly

astral303:
I noticed that traditional JavaDoc (`{@code}`) or Markdown-styled (```) examples embedded inside of comments would lose their formatting and indentation. 

This patch modifies the trimming to be just enough.


vojtechhabarta:
@astral303 thanks for your improvement.
It really seems that JavaDoc leaves just one space on lines (except the first).

PR #306, opened 2019-01-17
Support for emitting abstract methods found in beans

astral303:
* new setting to control this (off by default)
* supports JavaDoc enrichment, which is used to preserve parameter names
* Jackson1 and Jackson2 are supported
* avoids property setters and getters


vojtechhabarta:
I think this feature doesn't fit in intended support for mapping JSON classes (and REST resources) to TypeScript declarations.

Anyway thanks for your interest in typescript-generator.

PR #311, opened 2019-02-22
Added additional options for optional property declarations

rhandersen:
**Undefined use case:**
Properties in interfaces that can be undefined but are not optional to implement for classes implementing the interface. 

**Null or undefined use case:**
Same as the use case for undefined, but additionally we see a usage when doing partial resource updates undefined means "do not update property" whereas null means "set property to null in Java", i.e. deserialized to something like:
Optional<String> property = null;
Optional<String> property = Optional.ofNullable(null);

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/OptionalPropertiesDeclaration.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/OptionalPropertiesDeclaration.java
@@ -3,7 +3,9 @@ package cz.habarta.typescript.generator;
 
 
 public enum OptionalPropertiesDeclaration {
-    questionMark,
     nullableType,
+    nullableTypeAndUndefined,
+    questionMark,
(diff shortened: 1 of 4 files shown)


vojtechhabarta:
Thanks for the PR.

Could you please change the options to this:
``` java
public enum OptionalPropertiesDeclaration {
    questionMark,
    questionMarkAndNullableType,
    nullableType,
    nullableAndUndefinableType,
    undefinableType,
}
```

rhandersen:
> Thanks for the PR.
> 
> Could you please change the options to this:
> 
> ```java
> public enum OptionalPropertiesDeclaration {
>     questionMark,
>     questionMarkAndNullableType,
>     nullableType,
>     nullableAndUndefinableType,
>     undefinableType,
> }
> ```

Fixed. Thanks for the quick reply.

vojtechhabarta:
Thanks for the PR and quick change.

PR #323, opened 2019-04-02
Allow setting constructor on bean model

nivertius:
Constructor field in `TsBeanModel` looks like it could be changed (with clone-and-set), but there's method missing. This pull request adds the method.
This method is required for extension which adds constructor with required fields to the generated code.


vojtechhabarta:
Merged.
Usually I am adding methods as needed so thank you for submitting PR with method you (or somebody else might) need.

PR #324, opened 2019-04-02
Extension for generating constructors with required properties 

nivertius:
Here is an extension which generates constructors for generated classes. This constructor has a required parameter for each property that is `readonly` and has multiple possible values. This parameter is assigned to associated property. For each property which has a single possible value (as in union types with single string literal type), this value is assigned to the field.

This could be implemented as constructors which declare parameters as field (using any modifier), but as this is an extension, this would change already-generated bean classes by removing properties, which would be hard to do right.

This implementation currently doesn't support extension classes (class inheritance), because accessing superclass constructor declaration is not easy from the bean model, and it would require ordered bean processing, in order of inheritance (parent classes first). For now, this extensions throws exception when encounters class inheritance in any generated class.

Constructors are only generated for classes that have not defined constructor earlier (other extensions might do it).

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/RequiredPropertyConstructorExtension.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/RequiredPropertyConstructorExtension.java
@@ -1,0 +1,110 @@
+package cz.habarta.typescript.generator.ext;
+
+import cz.habarta.typescript.generator.Extension;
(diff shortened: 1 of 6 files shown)


vojtechhabarta:
Thanks for your extension! Merged.

vojtechhabarta:
I released [v2.13.489](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.13.489) with this feature.

nivertius:
Thank you very much!

PR #329, opened 2019-04-09
 Enable @JsonIdentityReference and @JsonIdentityInfo for bounded type parameters

panzerj:
It would be helpful if I could use @JsonIdentityInfo and @JsonIdentityReference anotations on a generic property at least if it has bounds like in the following example:

```
public class ValueAssignment<M extends IdData, T> {

    @JsonIdentityInfo(generator = ObjectIdGenerators.PropertyGenerator.class, property = "id")
    @JsonIdentityReference(alwaysAsId = true)
    private M idReference;

    private T assignedValue;
}
```


vojtechhabarta:
This looks nice.
If you don't need exact type of `M` and can just use `IdData` then this is good solution (and simple implementation).
Thanks.

vojtechhabarta:
Released in [2.14.505](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.14.505).

PR #336, opened 2019-04-29
Add whitelist for classes which should have constructor generated

nivertius:
This change adds possibility to specify which classes should have constructor generated. Configuration is provided as list of FQCNs, separated by whitespace, like that:
```xml
<extensionsWithConfiguration>
    <className>cz.habarta.typescript.generator.ext.RequiredPropertyConstructorExtension</className>
    <configuration>
        <classes>
            org.example.project.FirstClass
            org.example.project.SecondClass
        </classes>
    </configuration>
</extensionsWithConfiguration>
```

If no configuration parameter is provided, or extension is inserted not as `extensionWithConfiguration`, then all classes have constructor generated. This is the same behavior as current.

PR #337, opened 2019-04-29
Allow generating constructors for child classes

nivertius:
This removes the limitation of not being able to generate constructors if any bean has a parent class.

Solution doesn't remove limitations completely:
* Parent class constructor must still be generated with the same extension (cannot be preexisting). This limitation could be probably further removed to read any constructor, but this would require providing access to the parent `TsBeanModel`, which I assume is currently breaking model compilation layers.
* Bean must be inheriting from a reference type. Excluding some extreme cases, which I cannot even think of, this will be always true.

If bean does not inherit from any other class, constructor is generated as before. If bean inherits from parent, its constructor parameters are extracted and prepanded to current constructor parameters. Then, a superconstructor call is added as a first statement of a constructor, with arguments same as added parameters.

This has a problem of parent bean having the same parameter names as current bean properties. This is not currently handled in any way, but it will produce compilation error when compiling produced TypeScript code.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
@@ -48,6 +48,7 @@ public class ModelCompiler {
         BeforeTsModel,
         BeforeEnums,
         BeforeSymbolResolution,
+        AfterDeclarationSorting,
     }
 
     public TsModel javaToTypeScript(Model model) {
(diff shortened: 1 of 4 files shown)


vojtechhabarta:
@nivertius It seems there is a conflict after I merged your first PR. Could you try to solve it?

nivertius:
Done. :-)

vojtechhabarta:
Thanks

nivertius:
Could you please release 2.13.500, which contains this and #336? I would really appreciate it.

nivertius:
Wait. There's a bug in my code, check for existing constructor is wrong. I'll fix it right now.

nivertius:
No, the conflict resolution I did yesterday was bad, whitelist should be checked before creating constructor.

vojtechhabarta:
@nivertius this PR is already merged and I think I cannot do anything with it. You will probably need to submit new PR.

vojtechhabarta:
I see, you already submitted one (#339).
Merged. Thanks

vojtechhabarta:
Released in [2.14.505](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.14.505).

PR #342, opened 2019-05-02
 Use the BeanScanner directly instead of running the application

Antibrumm:
Hi there.
I was just going to write a similar tool and then i found yours :) Looks pretty decent already, but I think that running the full application to get the controllers one can loose a lot of time. 

This PR uses the Bean scanning feature of Spring to find Controllers without running the application.

### DIFF
--- a/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
+++ b/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
@@ -19,25 +19,20 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


vojtechhabarta:
@Antibrumm thanks for your interest in typescript-generator.

I am not sure whether directly using `BeanScanner` would behave in the same way as starting the application.

By the way instead of using `scanSpringApplication` parameter to find your controllers you can use for example `classPatterns` or `classesWithAnnotations` parameters so you don't have to start the application.

Antibrumm:
Hi, yes you are right. It will not use the `@Configuration` annotations to drill in but scan the "whole" classpath for `@Rest/Controllers`. 
My first try by using `classPatterns` was not working, but I just did a quick test and got similar results finally. 

The scan with `classesWithAnnotations` was resulting into too much output as it also writes out the Annotation itself or other spring controllers:

```export interface BasicErrorController extends AbstractErrorController {}
```export interface RestController extends Annotation {}

Best results i got with `classPatterns`, and as far as I can tell  this works as fast as the `BeanScanner` from Spring.

vojtechhabarta:
There are also `excludeClasses` and `excludeClassPatterns` parameters if you want to exclude some classes.

vojtechhabarta:
As I already mentioned `scanSpringApplication` is not only way how to find input classes so I think it is not needed to change this behavior now.

Antibrumm:
Thanks. I'm now using the classPattern feature to run some tests and up to now I don't see any problems with that approach.

Issue #348, opened 2019-05-09
NullPointerException when superclass is mapped to a primitive


Yona-Appletree:
Given a custom type mapping of a superclass to a primitive:

    // In build.gradle.kts, for example
    customTypeMappings = listOf(
        "org.test.SuperType:string"
    )

Or when using a `customTypeProcessor` that does the same.

typescript-generator will crash with a `NullPointerException` when it attempts to map a subclass (Use Kotlin to avoid these types of issues! ;):

    Caused by: java.lang.NullPointerException
        at cz.habarta.typescript.generator.compiler.ModelCompiler.mappedToClass(ModelCompiler.java:248)
        at cz.habarta.typescript.generator.compiler.ModelCompiler.processBean(ModelCompiler.java:193)
        at cz.habarta.typescript.generator.compiler.ModelCompiler.processModel(ModelCompiler.java:157)
        at cz.habarta.typescript.generator.compiler.ModelCompiler.javaToTypeScript(ModelCompiler.java:58)
        at cz.habarta.typescript.generator.TypeScriptGenerator.generateTypeScript(TypeScriptGenerator.java:67)
        at cz.habarta.typescript.generator.TypeScriptGenerator.generateTypeScript(TypeScriptGenerator.java:57)
        at cz.habarta.typescript.generator.gradle.GenerateTask.generate(GenerateTask.java:196)

This occurs because `getOriginClass(symbolTable, parentType)` can return `null` and `mappedToClass()` does not handle nulls. 

I'll submit a PR to fix this basic case, but some further thought might be given into handling of polymorphic custom type mappings.

vojtechhabarta:
@Yona-Appletree your fix for this problem is released in [2.15](https://github.com/vojtechhabarta/typescript-generator/releases).
Thanks again.

PR #349, opened 2019-05-09
#348: Fixed NPE when superclasses are mapping to primitives

Yona-Appletree:
Adds a test and a null-check for `mappedToClass` which I think should return false for a null `cls`.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
@@ -245,7 +245,7 @@ public class ModelCompiler {
     }
 
     private boolean mappedToClass(Class<?> cls) {
-        return !cls.isInterface() && settings.getMapClassesAsClassesFilter().test(cls.getName());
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Thanks for your PR.
I agree, for `null` input the method should return `false`.
Could you please just remove `System.out.println(output)` line?

Yona-Appletree:
I'll remove it, but I should point out that the `println` is there because I copied my tests from `cz.habarta.typescript.generator.CustomTypeMappingTest` where `testEnumAsMap` and `testSuperTypeString` both have such `println`s, so I thought it was convention for your tests.

vojtechhabarta:
It's not an intention to print the generated output in unit tests. I use this output when writing tests and I often forget to remove it. I just found it also in `testRegExpInPath()` (also my commit).

Just remove your line if you want to have small PR or remove it also from other places if you want to clean it up. It's up to you.
Thank you very much.

vojtechhabarta:
I merged your PR and then removed `println` lines from mentioned tests.
Thanks again.

vojtechhabarta:
Released in [2.15](https://github.com/vojtechhabarta/typescript-generator/releases).

Issue #353, opened 2019-05-22
Overriding Class names in generated typescript


peterklijn:
Hi,

I was wondering if there is a way to have the generated class have a different name than the class in Java, see the following example:

```java
public class Product {
  public String id;
  public Type type;
  ...

  public enum Type {
    SOME_TYPE, OTHER_TYPE
  }
}
```

When using this code in java, you would call the enum like `Product.Type.SOME_TYPE`. However the generated typescript flattens this structure:

```typescript
interface Product {
    id: string;
    type: Type;
}

enum Type {
    SOME_TYPE = "SOME_TYPE",
    OTHER_TYPE = "OTHER_TYPE",
}
```

This enum now looses this hierarchy, causing the name 'Type' to be too generic.

Renaming the enum to `ProductType` would be fine in the generated code, however in java calls like `Product.ProductType.SOME_TYPE` are a bit double in my opinion.

In Swagger, you have the ability to use different names, for example:

```java
public class Product {
  public String id;
  public Type type;
  ...

  @ApiModel(value = "ProductType")
  public enum Type {
    SOME_TYPE, OTHER_TYPE
  }
}
```

This would cause the generated code to be called 'ProductType', while leaving the original code uneffected.

Is there some kind of functionality like this in typescript-generator?

Thanks in advance.

vojtechhabarta:
Hi, yes, it is possible to rename types. It can be done using [customTypeNaming](http://www.habarta.cz/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html#customTypeNaming) _configuration_ parameter. For your example you can add `your.package.Product$Type:ProductType`.

If you have many of these cases you can also experiment with [customTypeNamingFunction](http://www.habarta.cz/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html#customTypeNamingFunction).

peterklijn:
Thanks a lot for your help @vojtechhabarta, this indeed solves the problem.

To be honest I found it a bit hard to figure out how to use this configuration parameter. Guessing by your documentation, it's generated from source code, so I made a PR to elaborate on the use of this [here](https://github.com/vojtechhabarta/typescript-generator/pull/356).

vojtechhabarta:
Thanks for your willingness to improve parameter description. In this case I would rather improved Wiki page here: https://github.com/vojtechhabarta/typescript-generator/wiki/Type-Mapping#classes-and-interfaces and just link it from parameter description.

[...]

PR #356, opened 2019-05-23
Added more elaborate explanation of customTypeNaming parameter

peterklijn:


vojtechhabarta:
I improved Wiki page (discussed in #353).
Thanks

PR #360, opened 2019-06-04
Improve eclipse developer experience

ruediste:
After checking out the sources and importing in the eclipse ide, I had the following issues:

- missing maven lifecycle mappings
- immutables apt processor not picked up by eclipse

below are the changes to the pom.xml which solve these issues. What do you think?


vojtechhabarta:
It is surprising that eclipse still doesn't have good support for Maven.

I would add `lifecycle-mapping` plugin if it solves the problem (hopefully it doesn't affect normal build).
But I don't want to add `immutables` dependency to `maven-compiler-plugin` which affects normal build. It is only used in test so maybe that's the reason why eclipse doesn't pick it up. Is there other solution? Or is it possible to use this project even without `immutables`?

ruediste:
`lifecycle-mapping`: Yes, that's just read by eclipse.

Maven Dependency: Just changed it to use `annotationProcessorPaths`. Works with eclipse. Better like that?

vojtechhabarta:
`annotationProcessorPaths` is even worse because it overrides default classpath for main build.

Also adding whole `lifecycle-mapping` section which is useful just for one IDE to the build doesn't look good to me.

I tried to open the project in eclipse and I must say it wasn't straightforward but on the other hand for experienced eclipse users some guide on what needs to be configured should be sufficient.

If you are interested we can create Wiki page with description how to open this project in different IDEs.

ruediste:
What is the big issue with changing the compilation classpath? Does this affect the generated .jar files?

vojtechhabarta:
Changing `maven-compiler-plugin` configuration affects both _main_ and _test_ jars. Typescript-generator main jar should not depend on Immutables library in any way. It is used just in tests. Immutables added to main jar build can generate new classes, change existing classes and it could also potentially fail the build because of some bug.

However it is fine to add it to eclipse build. If it could be possible to configure it outside pom.xml file in some eclipse specific file it would be fine.

ruediste:
I just did another run on this:

According to the documentation, `annotationProcessorPaths` should have no impact on the output:

> Classpath elements to supply as annotation processor path. If specified, the compiler will detect annotation processors only in those classpath elements. If omitted, the default classpath is used to detect annotation processors

Thus, as long as we do not use other annotation processors, no change should arise. I checked that by building `typescript-generator-core-2.16-SNAPSHOT.jar` with and without the `annotationProcessorPaths` configuration. The results were identical except for one build timestamp.

Thus I see no reason not to include this in the source code.

Issue #359, opened 2019-06-04
Customizeable TS Naming from Java


ruediste:
Hi

I have some specific wishes for the naming of generated TS classes (mainly regarding their namespace). I use typescript generator directly from java (not via maven)

I have seen the prefix/suffic settings and Settings.customTypeNamingFunction. However it feels a awkward to use some javascript for the type naming function when launching the generator from java. Would it make sense to add a setting taking an instance of cz.habarta.typescript.generator.compiler.SymbolTable.CustomTypeNamingFunction directly?

Regards

Btw: My current workaround:

	settings.extensions.add(new cz.habarta.typescript.generator.Extension() {
	
		@Override
		public EmitterExtensionFeatures getFeatures() {
			EmitterExtensionFeatures features = new EmitterExtensionFeatures();
			features.worksWithPackagesMappedToNamespaces = true;
			return features;
		}
	
		@Override
		public List<TransformerDefinition> getTransformers() {
			return Arrays.asList(new TransformerDefinition(ModelCompiler.TransformationPhase.BeforeSymbolResolution,
					new ModelTransformer() {
						@Override
						public TsModel transformModel(SymbolTable symbolTable, TsModel model) {
							for (TsBeanModel bean : model.getBeans()) {
								String name = bean.getOrigin().getName();
								String[] parts = StringUtils.removeStart(name, "ch.ecamos.").split("\\.");
	
								settings.customTypeNaming.put(name,
										parts[0] + "." + parts[parts.length - 1].replace('$', '.'));
							}
							return model;
						}
					}));
		}
	});

vojtechhabarta:
Hi,
typescript-generator should offer same set of features regardless how it is run (Maven, Gradle, directly from Java) so originally I didn't want to have this feature.

But indeed, it looks strange to write one function in JavaScript when everything else is written in Java. And I also think that suggested feature isn't something new, it is in fact existing feature just written in different language. So I will implement it.

vojtechhabarta:
Released in [2.15](https://github.com/vojtechhabarta/typescript-generator/releases).

PR #364, opened 2019-06-06
add Settings.customTypeNamingFunctionImpl

ruediste:
See #359

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -3,6 +3,7 @@ package cz.habarta.typescript.generator;
 
 import com.fasterxml.jackson.databind.Module;
 import cz.habarta.typescript.generator.compiler.ModelCompiler;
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
I was thinking about adding something like this
``` java
public SymbolTable.CustomTypeNamingFunction customTypeNamingJavaFunction = null;
```
to `Settings` class so it is as close as possible to `customTypeNamingFunction` parameter.
What do you think?

ruediste:
That was also my first shot. But I have seen the invocation `function.getName(cls.getName(), cls.getSimpleName())`, so I thought it might be better to just send the whole class to the function. Both are fine with me, your call.

vojtechhabarta:
I think using `SymbolTable.CustomTypeNamingFunction` guaranties that this feature stays the same in JavaScript and Java in the future.

ruediste:
Great. Adjusted.

vojtechhabarta:
Merged, thanks.

vojtechhabarta:
Released in [2.15](https://github.com/vojtechhabarta/typescript-generator/releases).

PR #370, opened 2019-06-13
Revert ResponseEntity handling from #396

svpace:


vojtechhabarta:
Thanks for the quick fix and adding test for this case.

vojtechhabarta:
Released in [v2.15.527](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.15.527).

Issue #377, opened 2019-06-25
Add support for prefixing urls generated by rest client generation with an arbitrary string


mikeporterdev:
Currently when generating a client (in my case a spring rest controller) it will use the RestController path e.g.
```
@RestController
@RequestMapping("/foo")
public class Bar {
    @RequestMapping("/baz")
    public void getBaz() {
    ...
    }
}
```

Will generate 
```
return this.httpClient.request(... url: uriEncoding`foo/baz`...);
```

However due to our microservices architecture using Zuul proxy, all of our requests from the front end to back end need to go through `api/microserviceName/foo/baz` first, and Zuul then navigates the request to `microserviceUrl/foo/baz`. This api prefix is NOT on the RequestMapping of any of the controllers.

Because of this, it would be useful if we can prefix an arbitrary string to all generated client URLs.

Raising this issue to explain the use case for the PR I am about to raise as I've currently had to implement a bit of a hack in our front end by building one HttpClientAdapter for every one of our microservices. 

PS. Outstanding project, begun implementing this today and so far it is working like a dream.

mikeporterdev:
Similarly it would be useful to prefix the RestApplicationClient class, as multiple microservices would have a client each. But I can get around this in the meantime using package names so I haven't submitted a PR for this one. Happy to do so if you agree it's useful for general use though.

vojtechhabarta:
Thanks for your interest in this project and description of your feature.
I think it is not needed to add API prefix to all generated methods. Instead it can be set just once as a base URL. You can implement `HttpClient` interface and pass desired URL in constructor like this:
``` typescript
const client = new MyHttpClient("https://myserver.mydomain.com/api/microserviceName");
```
What do you think?
You can also try `cz.habarta.typescript.generator.ext.AxiosClientExtension` extension and see how it is done.

PR #378, opened 2019-06-25
Added support for prefixing the urls generated by Jaxrs/Spring client generation

mikeporterdev:
Fixes issue #377


vojtechhabarta:
I think it is not needed to add extra parameter.
It is possible to build needed URLs with current version, see https://github.com/vojtechhabarta/typescript-generator/issues/377#issuecomment-507067206.

PR #383, opened 2019-07-05
Create tagged type unions with type parameters when needed

RelikDream:
Generated tagged type union of type paramatrized types had no type parameters. This PR fix this.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
@@ -50,6 +50,7 @@ import cz.habarta.typescript.generator.parser.RestMethodModel;
 import cz.habarta.typescript.generator.parser.RestQueryParam;
 import cz.habarta.typescript.generator.util.Pair;
 import cz.habarta.typescript.generator.util.Utils;
+
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Thanks for this PR fixing the problem, typescript-generator really produces invalid TypeScript code in this case and it should be fixed.

But I think it would be really complex to handle all cases involving generics correctly.

Let's say we have following classes connected using `@JsonSubTypes` annotation:
``` java
class Vehicle<M extends Engine> {}
class Boat<M extends Engine> extends Vehicle<M> {}
class Plane<M extends Engine> extends Vehicle<M> {}
```

We can validate that sub-classes have same number of generic parameters as base class (as you already did) but then somebody can introduce _unrelated_ generic parameter like this:
``` java
class Truck<B extends BodyType> extends Vehicle<DieselEngine> {}
```

Or somebody can have classes with two or more generic parameters:
``` java
class Base<T1,T2,T3> {}
class DerivedA<T3,T2,T1> extends Vehicle<T1,T2,T3> {}
class DerivedB<A,B,C> extends Vehicle<C,A,B> {}
```
and we would need to determine correct order of generic parameters when generating union type.

So I think it's not worth it to generate union types for generic classes because added value is not so high. Users can consume those class hierarchies without union types.

I can implement it. It would behave as `disableTaggedUnions` parameter turned on for generic classes. What do you think?

RelikDream:
I think it's a pretty good idea.

vojtechhabarta:
@RelikDream I pushed commit that prevents generating tagged unions for generic classes. I used your test classes in unit test.
Thank you for identifying and helping to fix the issue.

vojtechhabarta:
@RelikDream I realized that tagged unions with generic parameters can actually be useful, see example in #414. In that PR I implemented support for this feature similarly to your PR but I also added check how **generic parameters** in descendants **relate** to generic parameters in abstract base class (as I mentioned in my previous comment). For this check I partially reused `GenericsResolver` class which I added for other use cases.

Thanks again for your work.

PR #388, opened 2019-07-22
New extension: OnePossiblePropertyValueAssigningExtension

krzs:
If you have a class hierarchy usually you add some discriminator property to distinguish between the subclasses (in JSON mapping). However, in these subclasses that properties can have only one value (usually the name of the class). This extension marks such properties (with only one possible value) as read only and sets their value in the constructor, so it will not be necessary to set it manually.


vojtechhabarta:
@krzs thanks for this extension.

PR #389, opened 2019-07-23
Add support generic bounds

### LINKED ISSUE #380: Generic type parameters not resolved properly
Given to following class hierarchy:

```java
public interface IdView<T extends Serializable> {
    T getId();
}

public interface Entity1IdView extends IdView<UUID> {
}

public abstract class Entity1View implements Entity1IdView {
  private String name;

  public String getName() { return name; }
}
```

Yields the following typescript definitions:

```typescript
export interface Serializable {
}
export interface Entity1IdView extends IdView<string> {
}
export class Entity1View implements Entity1IdView {
    id: T;
    name: string;
}
```

Apparently, the type parameter `T` is not resolved correctly for the declaration of the `id` field in `Entity1View`.

olegshtch:
Allows to use generate extends bound on generic types.

Suppose to fix #380


vojtechhabarta:
@olegshtch thanks for this nice PR.
I would wait for more use cases or comments before merging it.

olegshtch:
Looks like #380 isn't fixed with this PR.

olegshtch:
Rebased. Removed test for #380.

vojtechhabarta:
@olegshtch I pushed fix for #380 - c23129914fb5b096a4541c4ca26caf6804513643

olegshtch:
@vojtechhabarta I suppose generic bounds wasn't part of this commit?

vojtechhabarta:
No, I think it is separate thing.

olegshtch:
It something left to do in this PR?

vojtechhabarta:
I am not sure if this feature has high enough value to justify the change in generated output.
Do you have any real example where this feature might be useful?
The feature itself is nice but I think it can potentially break some existing user code.

olegshtch:
We use a narrowing of types in a production Java code so prefer to have it in a TypeScript. Something like `Message<T extends Body>` where `Body` have a some guarantied fields.

olegshtch:
Maybe add a new option to enable it?

vojtechhabarta:
In general I prefer not to add new parameters because it is becoming harder for users to go through all typescript-generator parameters.

Currently typescript-generator outputs `Message<T>` which looses `extends Body` but then this type is used somewhere with particular type argument, for example:
``` typescript
interface Response {
    message: Message<HumanBody>;
}
interface HumanBody extends Body {
}
```
So when there is a JSON with `Response` data it is correctly typed.

olegshtch:
Rebased

PR #396, opened 2019-07-26
Order optional parameters to constructor as last

### LINKED ISSUE #395: RequiredPropertyConstructorExtension doesn't order constructor fields correctly when OptionalAnnotations is used.
Given two Java classes
```java
public class TestSuper {
    private String requiredSuperField;
    @Nullable
    private String optionalSuperField;
}

public class TestSubClass extends TestSuper {
    private String requiredSubField;
}
```

The following output is given

```typescipt
export class TestSuper {
    readonly requiredSuperField: string;
    readonly optionalSuperField?: string;

    constructor(requiredSuperField: string, optionalSuperField?: string) {
        this.requiredSuperField = requiredSuperField;
        this.optionalSuperField = optionalSuperField;
    }
}

export class TestSubClass extends TestSuper {
    readonly requiredSubField: string;

    constructor(requiredSuperField: string, optionalSuperField?: string, requiredSubField: string) {
        super(requiredSuperField, optionalSuperField);
        this.requiredSubField = requiredSubField;
    }
}
```

This gives the following typescript compiler error:
error TS1016: A required parameter cannot follow an optional parameter. on the requiredSubField parameter to the constructor.

Maven settings.
```xml
ions>
    <extension>cz.habarta.typescript.generator.ext.RequiredPropertyConstructorExtension</extension>
</extensions>
<declarePropertiesAsReadOnly>true</declarePropertiesAsReadOnly>
<optionalAnnotations>
    <annotation>javax.annotation.Nullable</annotation>
</optionalAnnotations>
```

nivertius:
Optional, but readonly fields in classes become optional parameters in constructor. These don't work well with inheritance without any modifications, because optional parameters can't be in constructor before non-optional ones.

This change fixes the problem by pushing optional parameters after required ones. Parameters will still retain their order in respect to position in hierarchy and declaration within class, i.e. required base class parameters will be positioned in constructor before required subclass parameters, and required parameters that are declared earlier will be before required parameters declared later. Same is true for pairs of optional parameters. But required parameters will be always before optional.

Fixes #395

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/RequiredPropertyConstructorExtension.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ext/RequiredPropertyConstructorExtension.java
@@ -92,6 +92,7 @@ public class RequiredPropertyConstructorExtension extends Extension {
     private static Optional<TsConstructorModel> createConstructor(TsBeanModel bean, TsModel model,
                                                                   Map<String, TsConstructorModel> generatedConstructors) {
         List<TsParameterModel> parameters = new ArrayList<>();
+        List<TsParameterModel> optionalParameters = new ArrayList<>();
         List<TsStatement> body = new ArrayList<>();
         TsType parent = bean.getParent();
         if (parent != null) {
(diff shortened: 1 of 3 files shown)


vojtechhabarta:
@nivertius thank you for this PR

PR #405, opened 2019-08-16
Update & fixes for Checkstyle

danielkaneider:
Previously Eclipse showed a bunch of CS violations. In order to tackle this following has been changed:

- Updated the Maven CS plugin to 3.1 (which is now based on CS 8.19 instead of CS 6.1)
- The build now fails on CS violations
- Fixed some files with CS violations

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,11 +105,8 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <version>3.1.0</version>
                 <configuration>
-                    <includeTestResources>true</includeTestResources>
                     <includeTestSourceDirectory>true</includeTestSourceDirectory>
                     <configLocation>../checkstyle.xml</configLocation>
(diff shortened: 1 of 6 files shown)


vojtechhabarta:
Thanks for this PR.

What's the purpose of `includeTestResources`, `consoleOutput`, and `failsOnError` parameters? I think current configuration works well (also with maven-checkstyle-plugin 3.1.0). It fails build on errors and I can also see violations on output. I would prefer to leave the configuration as is (if this is suitable for you as well).

danielkaneider:
You were right. `mvn install` seems to fail also without those properties set, so I removed them

vojtechhabarta:
Thanks

vojtechhabarta:
Released in version [2.17.558](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.17.558).

PR #406, opened 2019-08-16
Smaller refactorings

danielkaneider:
This PR contains some refactorings done in #237 .

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.stream.Collectors;
 
 
 /**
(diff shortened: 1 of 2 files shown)


typescript-generator-core/src/main/java/cz/habarta/typescript/generator/TsType.java
```diff
@@ -49,6 +50,12 @@ public int hashCode() {
     public String toString() {
         return format(new Settings());
     }
+    
+    public static List<String> strinfigyList(List<? extends TsType> types, Settings settings) {
```
vojtechhabarta: I think this method is not needed, you can use `Emitter.formatList` methods.
Or if it is really needed in your case could you please rename it to `formatList`?

danielkaneider: It's not needed; fixed.
Please consider changing `Utils.join` to just accept Iterables of `String` instead of `Object`. This would prevent potential issues.

vojtechhabarta: Thanks.
`Utils.join` was useful in Java 7. I will remove it completely because from Java 8 it can be replaced by `String.join` or `Collectors.joining`.

PR #407, opened 2019-08-16
gson: add support for @SerializedName

ruediste:
Add support for the GSON @SerializedName annotation.

This changes the property names in the generated typescript code. Is this how the other libraries handle things?

### DIFF
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -62,6 +62,11 @@
             <artifactId>classgraph</artifactId>
             <version>4.8.24</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


vojtechhabarta:
Thanks for this PR.

Just to answer your question: for Jackson library it is done differently - list of properties is obtained from Jackson library which processes not only `@JsonProperty` annotation but also `@JsonIgnore` and many others.

So if you would find a way how to get list of properties for specified class from Gson library it would be great. But processing few annotations directly is also fine.

vojtechhabarta:
Released in version [2.17.558](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.17.558).

PR #416, opened 2019-09-18
custom annotations for Spring Controllers

AMilassin:
# Why
The goal with this PR is to support filtering Spring Controllers processed for client generation with Java annotations. I think in some cases this is easier than filtering classes in the configuration.

It also enables using interface-based separation of definitions. Here adding `@RestController` would be superfluous.

E.g.:
```
     @GenTypeScript
     @RequestMapping("/credentials")
     public class LoginApi {

          @GetMapping("/login")
          boolean login(@RequestBody String loginName, @RequestBody String password);
     }

     @RestController
     public class LoginController implement LoginApi {
         @Override
         public boolean login(String loginName, String password) {
             // ...
         }
     }
```

# Implementation
The default implementation was triggered for each controller with `@RestController` present. This change provides a configuration to change the marker annotation which triggers the client generation of a Spring Controller. New setting is named `springControllerAnnotation `.

Only implemented for Spring client generation.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -81,6 +81,7 @@ public class Settings {
     public boolean generateSpringApplicationInterface = false;
     public boolean generateSpringApplicationClient = false;
     public boolean scanSpringApplication;
+    public String springControllerAnnotation;
     @Deprecated public RestNamespacing jaxrsNamespacing;
     @Deprecated public Class<? extends Annotation> jaxrsNamespacingAnnotation = null;
     @Deprecated public String jaxrsNamespacingAnnotationElement;  // default is "value"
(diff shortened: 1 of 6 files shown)


vojtechhabarta:
@AMilassin thanks for your interest in improving typescript-generator.

There is `classesWithAnnotations` parameter which should be also usable for Spring controllers. Could you please try it if it works in your case? (Leave `scanSpringApplication` parameter `false`.) Maybe I should just improve description to make it more discoverable for users.

AMilassin:
Thanks for your quick response @vojtechhabarta .

You are right. I think we have an edge case as not counting typescript-generator we would not put `@RestController` on our Spring Controller interfaces. We have separate api declarations as interface files and separate implementation classes. The api interfaces doesn't not require the `@RestController` annotation.

From this perspective this seems like a less important product feature.

I still think that limiting the controller detection to a certain annotation causes edge cases to fail. Like using `@Controller` or a custom annotation would. The possibility to customize the annotation would alleviate this problem.

vojtechhabarta:
I was mistaken. I thought the issue is about finding those controller classes however the issue is that typescript-generator knows about them but doesn't treat them as controllers when they do not have `@RestController` annotation.

In your case I would generate TypeScript from `LoginController` because it represents real REST service but I think I understand your motivation.

Do you think this issue could be solved without adding new configuration parameter? Maybe you could have some annotation "derived" from `@RestController` that could be used for finding your API controllers and it could also be recognized as controllers. For example:
``` java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@RestController
@interface ApiRestController {
}
```

AMilassin:
Thanks @vojtechhabarta this is indeed the correct solution here!

PR #435, opened 2019-10-28
Add support for custom mappings of supertypes

headphonejack:
All subtypes that are applicable to the supertype will be mapped using the custom mapping.

Looking for feedback and design review.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/CustomMappingScope.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/CustomMappingScope.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (C) 2019 - present by OpenGamma Inc. and the OpenGamma group of companies
+ *
+ * Please see distribution for license.
+ */
+package cz.habarta.typescript.generator;
+
+public enum CustomMappingScope {
+
+    /**
+     * A mapping of a concrete class
+     */
+    CLASS,
+
+    /**
+     * A mapping of a super type, such as an abstract class or interface.
+     */
+    SUPERTYPE,
+}
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/CustomMappingScope.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/CustomMappingScope.java
@@ -3,12 +3,12 @@ package cz.habarta.typescript.generator;
 public enum CustomMappingScope {
 
     /**
-     * A mapping of a concrete class
+     * A mapping of a concrete class.
      */
     CLASS,
 
     /**
-     * A mapping of a super type, such as an abstract class or interface.
+     * A mapping of a supertype, such as an abstract class or interface.
      */
     SUPERTYPE,
 }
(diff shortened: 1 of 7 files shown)


headphonejack:
@vojtechhabarta looking for someone to take a look at this, but not able to assign anyone myself

vojtechhabarta:
@headphonejack thanks for your pull request.

Custom type mappings feature supports generics so this new feature should also supports generics to be consistent. But it can be tricky. I am thinking about following example:
``` java
class BaseClass<A, B> {}

class SubClass1<B> extends BaseClass<String, B> {}

class SubClass3<X, Y, Z> extends BaseClass<Z, Y> {}

class MyClass {
    public SubClass1<Integer> property1;
    public SubClass3<String, Boolean, Integer> property3;
}
```
then I could have mapping like `? extends package.BaseClass<A,B>:TypeScriptBaseClass<T,U>`
what should be generated in this case? Probably following:
``` typescript
interface MyClass {
    property1: TypeScriptBaseClass<string, number>;
    property3: TypeScriptBaseClass<number, boolean>;
}
```
There is `cz.habarta.typescript.generator.util.GenericsResolver` utility class which could be used for this.

What do you think?

headphonejack:
Thanks @vojtechhabarta - I think that it sounds reasonable to attempt to support generics as the other custom mappings do - I will take a look if there are any gotchas

vojtechhabarta:
Thanks for recent improvements. I would like to ask for few changes before merging this PR. Fist thing is syntax of the mapping. For me `? extends BaseClass` was confusing because we are not configuring mapping of bounded wildcard type (`? extends BaseClass`) but we are configuring mapping of some class and its descendants. So I would prefer different syntax. I don't have many proposals, maybe just `+` character before (or after) the class name. So mapping would look like `+package.BaseCustomMapping:string`.

I would also change few things in implementation:
- I would load supertype classes early in `getValidatedCustomTypeMappings()` and set loaded class to `CustomTypeMapping` object. This should eliminate need to propagate classloader into `CustomMappingTypeProcessor` through `TypeProcessor.Context` and `SymbolTable`.
- Instead of `CustomMappingScope` enum I would use just `boolean` property.
- I would inline private constants like `JAVA_NAME_REGEX_GROUP` and `SUPERTYPE_NAME_PATTERN` so it is easier to read code without jumping between two places.

So `CustomTypeMapping` class could look like this:
``` java
public static class CustomTypeMapping {
    public final GenericName javaType;
    public final GenericName tsType;
    public final boolean isSuperclassMapping;
    public final Class<?> superclass;

    public CustomTypeMapping(GenericName javaType, GenericName tsType) {
        this.javaType = javaType;
        this.tsType = tsType;
        this.isSuperclassMapping = false;
        this.superclass = null;
    }

    public CustomTypeMapping(GenericName javaType, GenericName tsType, Class<?> superclass) {
        this.javaType = Objects.requireNonNull(javaType);
        this.tsType = Objects.requireNonNull(tsType);
        this.isSuperclassMapping = true;
        this.superclass = Objects.requireNonNull(superclass);
    }
}
```
First constructor is partially for backwards compatibility.

Does it make sense to you?

vojtechhabarta:
@headphonejack in #545 I implemented general support for data libraries. I needed to add support for superclass mapping with generics resolving. Your PR and our discussion was very helpful while implementing it. Thanks.

Even though `customTypeMappings` parameter still doesn't support superclass mappings the logic for it is already implemented so I can add this support if anybody is interested.

Issue #421, opened 2019-10-03
Map from enum


RobbinBaauw:
Currently, as far as I could find in the codebase, any map is turned into `{ [index: string]: ...}`, as visible here:
https://github.com/vojtechhabarta/typescript-generator/blob/48b30bd76998c8e35f99a727801566b7fabe730d/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/DefaultTypeProcessor.java#L84-L87

But it would be nice if a `Map<SomeEnum, Any>` could get mapped into `{ [index in keyof typeof SomeEnum]: ...}`. This would allow you to do typesafe accesses.

**Edit**: This seems to be fully possible (using enums mapped to a custom string value, mapped to union type and mapped to key name).

Example 1:
```java
enum Test {
    TEST1,
    TEST2
}

Map<Test, String> x
```
maps to
```typescript
enum Test {
    TEST1 = "TEST1",
    TEST2 = "TEST2"
}

{ [index in Test]: string}

// Could be initialized as
... = {
   TEST1: "somethingTEST1",
   TEST2: "somethingTEST2",
}
```

Example 2:
```java
enum Test {
    TEST1("a"),
    TEST2("b")

    @JsonValue
    private String value;

    ...constructor
}

Map<Test, String> x
```
maps to
```typescript
enum Test {
    TEST1 = "a",
    TEST2 = "b"
}

{ [index in Test]: string}

// Could be initialized as
... = {
   a: "somethingA",
   b: "somethingB",
}
```

Example 3:
```java
enum Test {
    TEST1("a"),
    TEST2("b")

    @JsonValue
    private String value;

    ...constructor
}

Map<Test, String> x
```
maps to
```typescript
type Test ="a" | "b";

{ [index in Test]: string}

// Could be initialized as
... = {
   a: "somethingA",
   b: "somethingB",
}
```

In the future this could become even better: https://github.com/microsoft/TypeScript/pull/26797.

Is this a feature that could be in the library @vojtechhabarta? I'm willing to implement it

vojtechhabarta:
Thanks for this suggestion. It looks good but I think `{ [index in Test]: string }` type means object with *all* properties from `Test`. So this example
``` typescript
type Test = "a" | "b";

const x: { [index in Test]: string } = {
    a: "somethingA",
    b: "somethingB",
}
```
works well but
``` typescript
const y: { [index in Test]: string } = {
    a: "somethingA",
}
```
[...]

PR #438, opened 2019-11-10
Improve Spring support

### LINKED ISSUE #437: Merging JAX-Rs and Spring functionality
Hi,

This library has excellent JAX-Rs support, while I found that there are more problems when using Spring (e.g. resolving generics and more). I was wondering why this is, as it seems to me that both ways of REST controllers are pretty similar, though with a bit different annotation names and ways of using them.

You said in another issue that:
> Generics are resolved differently for JAX-RS and Spring - JAX-RS is resolved by typescript-generator itself while Spring is resolved using Spring libraries.

But can't we change that? It seems to me that the JAX-Rs parser can be mostly reused for Spring as well, while adding specifics for both in the subclass for the respective library (e.g. Pageable for Spring).

Maybe I'm completely wrong, but in my view it would help in a way that a fix or improvement would improve both Spring and JAX-Rs instead of just a single one.

If help is needed I'm willing to help :) If this seems like a possible idea I can create a POC.

RobbinBaauw:
This will improve Spring support a lot. Closes #437.

The following cases are improved:
- Generics resolving in both the return type and method arguments. I did this by switching from the Spring resolver to Typescript Generator's own generic resolver. This also fixed the fact that enum mapping didn't yet work (#421)
- `PathVariable` being string when used without value. In Spring it is possible to use `PathVariable` without a value if the URI part is the same as the variable name. Now this wasn't checked, leading to `PathVariable`s which were `string` (default) instead of the required type
- Inheritance giving double methods (could be useful for JAX-Rs too), I did this by looking in the inheritance chain and looking which methods override other methods and add a `RequestMapping` annotation. If they do that, use it, else use the method of the child type.

### DIFF
--- a/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
+++ b/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
@@ -1,8 +1,6 @@
 
 package cz.habarta.typescript.generator.spring;
 
-import static cz.habarta.typescript.generator.util.Utils.getInheritanceChain;
-
 import cz.habarta.typescript.generator.Settings;
 import cz.habarta.typescript.generator.TsType;
 import cz.habarta.typescript.generator.TypeProcessor;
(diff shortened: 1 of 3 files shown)


typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
```diff
@@ -75,7 +74,9 @@ public RestApplicationParser create(Settings settings, TypeProcessor commonTypeP
             return new SpringApplicationParser(settings, commonTypeProcessor);
         }
 
-    };
+    }
+
+    ;
```
vojtechhabarta: 🙂

vojtechhabarta:
Thank you very much for this pull request and description of changes.
I went briefly through changes and it looks very useful, Spring support in typescript-generator will be more consistent with Spring behavior.

typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
RobbinBaauw: Interesting... fixed now

PR #441, opened 2019-11-21
Add support for required @RequestParam

RobbinBaauw:
In Spring, `@RequestParams` are required by default, but this wasn't enforced by Typescript-Generator. Now it uses both `required` and `defaultValue` in order to determine if it is required.

I implemented it in such a way that if 1 query parameter is actually required, the entire `queryParams` object would be required (but individual params can be optional). If no query parameter is required, the entire object can be `undefined`.

This could be breaking, but this does correspond to Spring behaviour:
https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html


vojtechhabarta:
Thank you for this PR and sorry for late response.
I think the breaking change regarding required `queryParams` is worth it. Let's do it rather sooner than later. (Definitely I would not want to add configuration parameter for retaining original behavior.)

PR #442, opened 2019-12-03
Nullability

### LINKED ISSUE #304: Nullable collection elements
Hey there!

I'm using Kotlin to create all my DTOs. In Kotlin, you can do the following:

```kotlin
data class MyDto(
  val arrWithElements: Array<Int>,
  val arrWithNullableElements: Array<Int?>
)
```

The information doesn't exist at runtime unfortunately, as it's wiped away, therefore there's no easy direct way to generate an interface in TS that would look like this:

```typescript
interface MyDto {
  arrWithElements: number[]; // or Array<number>
  arrWithNullableElements: Array<number | null>;
}
```

What I'm proposing then, is the ability to configure a custom annotation in the generator, which if defined on a property that happens to be a collection, that collection would be understood to have nullable elements.

Please let me know what you think

### LINKED ISSUE #419: JetBrains @Nullable annotation support
Hi,

This is the same request as #173.

Being: would it be possible to flag `@org.jetbrains.annotations.Nullable` as an optional JSON annotation? When putting it in the `optionalAnnotations` it doesn't work (probably because of the `RetentionPolicy.CLASS`).

Thanks!

RobbinBaauw:
This adds nullability support for Kotlin.

It does that by using kotlin reflection and propagating `KType` together with `Type` throughout the library. This is because when processing the types in a `TypeProcessor` it is required to know what the `KType` is.

I used `OptionalType` and made it work with generics and the `OptionalPropertiesDeclaration` setting in the following way (see https://github.com/vojtechhabarta/typescript-generator/pull/442/files#diff-8e6a80fd6c309e32cff93310315f1092R1005), take for example `Array<String?>`:
- If `nullableType`: `(string | null)[]`
- If `questionMarkAndNullableType`: `(string | null | undefined)[]`. The `?` will be there when there are no generics, then the union type with `undefined` won't be there.
- If `nullableAndUndefinableType`:  `(string | null | undefined)[]`
- If `undefinableType`: `(string | undefined)[]`
- If `questionMark`: `(string | undefined)[]`. The `?` will be there when there are no generics, then the union type with `undefined` won't be there.
- If not set: `(string | undefined)[]`. The `?` will be there when there are no generics, then the union type with `undefined` won't be there (which was the behavior that was supported already)

This is tested in `OptionalTest` (https://github.com/vojtechhabarta/typescript-generator/pull/442/files#diff-9bedabc19734567ef9b2e18a3425dc20R106), where it also tests the `Optional<String>[]` behavior mentioned in #304.

For JAX-Rs and Spring it also works. I added tests for JAX-Rs in `KotlinTest` and for Spring in `SpringKotlinTest`.

I'm not sure if this is what you had in mind, but I needed it so I implemented it. 

Closes #419 
Closes #304 

Note that I built it upon #441


vojtechhabarta:
Thanks for this pull request.

It's relatively complex feature so I needed to implement it myself to understand it well. I used your PR as a prototype, I used your tests (very important) and your know how (I always checked how you did something).

Main difference is that I didn't want to propagate `KType` throughout the tool. So I parse nullability information and use internal model to represent it. This model is based on `java.lang.reflect.Type` which is not ideal but that's how it currently works. In addition to standard types there was already `JUnionType` and now also `JTypeWithNullability`. Later this model also allowed to represent nullability specified in Java classes using type annotations.

Second thing is that I added `TsType.NullableType` instead of reusing `TsType.OptionalType`. In fact I regret introducing `OptionalType` in the past because in TypeScript optionality is really attribute of properties and parameters, not a type. So it could be removed sometime in the future and represented correctly.

Thank you again for your very useful work and it would be great if you could check my implementation if it does what you need.

RobbinBaauw:
Closing because of #448

PR #448, opened 2020-01-07
Nullability in Kotlin and Java

vojtechhabarta:
This is continuation of #442.

### Nullable types in TypeScript

In _strict_ mode TypeScript supports [non-nullable types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#null--and-undefined-aware-types). This means that for example `string` type doesn't permit `null` and `undefined` values. If some variable should be able to store those values its type could be union `string | null | undefined`.

Example with arrays:
``` typescript
const list1: string[] = ["text", null];  // error
const list2: (string | null)[] = ["text", null];  // ok
```

### Nullable types in Kotlin

In Kotlin regular types don't permit `null` values. Nullable types can be created using `?` character.

Example with lists:
``` kotlin
val list1: List<String> = listOf("text", null)  // error
val list2: List<String?> = listOf("text", null)  // ok
```

### Nullable types in Java

Java doesn't have concept of nullability. But in Java 8 (and later) it is possible to use annotations also on types. So for example list which allows `null` values can be written as `List<@Nullable String>`. Please note that such nullable annotation must have `TYPE_USE` target. So it is possible to use for example `org.checkerframework.checker.nullness.qual.Nullable` but not `javax.annotation.Nullable`.

Example with lists:
``` java
List<@NonNull String> list1 = Arrays.asList("text", null);  // can be reported
List<@Nullable String> list2 = Arrays.asList("text", null);  // ok
```

## Typescript-generator's support for nullability

Typescript-generator is now able to represent internally and emit nullable types. For Kotlin classes nullability is detected automatically using Kotlin reflection. For Java classes it is needed to configure used nullable annotation(s) using `nullableAnnotations` parameter.

How exactly will nullable types look like in generated output is specified using `nullabilityDefinition` parameter. Let's say we have following Java class:
``` java
public class Lists {
    public List<String> list1;
    public @Nullable List<String> list2;
    public List<@Nullable String> list3;
    public @Nullable List<@Nullable String> list4;
}
```
When `nullabilityDefinition` parameter is set to `nullAndUndefinedUnion` output will be:
``` typescript
interface Lists {
    list1: string[];
    list2?: Nullable<string[]>;
    list3: Nullable<string>[];
    list4?: Nullable<Nullable<string>[]>;
}
type Nullable<T> = T | null | undefined;
```
Or when `nullabilityDefinition` parameter is set to `nullInlineUnion` (default) output will be:
``` typescript
interface Lists {
    list1: string[];
    list2: string[] | null;
    list3: (string | null)[];
    list4: (string | null)[] | null;
}
```


RobbinBaauw:
I think this looks good! Maybe the default of `nullAndUndefinedUnion` is not the best, as people might expect Java / Kotlin `null` to be the same in TypeScript, which it isn't with this setting. Maybe `nullUnion` would be better.

vojtechhabarta:
Thanks for quick response.

I was thinking that including `undefined` in nullable types by default is more "safe" because Jackson can skip `null` properties (leaving them undefined) when serializing object into JSON. For example:
``` java
public static class A {
    public String text = null;
    public List<String> list = Arrays.asList("item", null);
}

ObjectMapper objectMapper = new ObjectMapper();
objectMapper.setDefaultPropertyInclusion(JsonInclude.Include.NON_NULL);
System.out.println(objectMapper.writeValueAsString(new A()));
```
outputs JSON without `text` property:
``` json
{
    "list": [
        "item",
        null
    ]
}
```
And in opposite direction when Jackson is deserializing JSON it doesn't matter if JSON property is `null` or missing.

On the other hand default Jacksons behavior is to write `null` properties. Also arrays in JSON cannot contain `undefined` values.

So I am still not sure about the default but I will change it. Do you prefer `nullUnion` or `nullInlineUnion`?

RobbinBaauw:
Good! I think sticking to the Jackson default is indeed best. I'd prefer `nullInlineUnion`, as you're only 1 click away from viewing the type instead of 2. 

Note that I didn't check the entire PR, I mostly checked whether the reason that I propagated the types through the chain (because of nested types) was handled, and it was. Looking forward to a new release with this!

vojtechhabarta:
Default is changed.
It's ok that you checked just what's interesting to you, I will release it and we will see.

RobbinBaauw:
Hi, could you release this version so I don't have to use my own patched version anymore?

vojtechhabarta:
Released [2.19.575](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.19.575).

vojtechhabarta:
Released [2.19.577](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.19.577).

PR #465, opened 2020-03-23
adding JSON-B support

rmannibucau:
High inspired from Apache Johnzon, this PR adds JSON-B support to typescript-generator.

### DIFF
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -77,6 +77,11 @@
             <artifactId>kotlin-reflect</artifactId>
             <version>${kotlin.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.geronimo.specs</groupId>
+            <artifactId>geronimo-jsonb_1.0_spec</artifactId>
+            <version>1.2</version>
+        </dependency>
         <!--test dependencies-->
         <dependency>
             <groupId>junit</groupId>
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/JsonLibrary.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/JsonLibrary.java
@@ -1,5 +1,5 @@
 package cz.habarta.typescript.generator;
 
 public enum JsonLibrary {
-    jackson1, jackson2, jaxb, gson
+    jackson1, jackson2, jaxb, gson, jsonb
 }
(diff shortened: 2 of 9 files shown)


vojtechhabarta:
Thanks for this great PR and sorry for late response.

I think `jakarta.json.bind:jakarta.json.bind-api` is more "official" JSONB API dependency.
I will change it.

rmannibucau:
@vojtechhabarta oki, np. The main difference is about the license (ASFv2 vs EPL) but guess both are ok for this project. Thanks for merging :).

Side question: any idea when it can be released (no pression, just to get a rough idea)?

vojtechhabarta:
Usually I plan to do release once a month.

vojtechhabarta:
Released in [v2.22.595](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.22.595). Thanks!

vojtechhabarta:
@rmannibucau just for your information: I was cleaning the code and I fixed also some compiler warnings in `JsonbParser`. But more importantly I changed which class loader is used for loading classes. Parsers need to use class loader from `Settings` object which is correctly assembled by build plugins (typescript-generator-maven-plugin and typescript-generator-gradle-plugin).

I also have a question regarding `JohnzonAny`. This is annotation from Johnzon library so strictly speaking it should not be handled by generic JSONB parser. Do you think it is really needed? And do you think we will need to add support for other Johnzon annotations in the future?

rmannibucau:
@vojtechhabarta ok for the classloader/refactoring. About johnzon the question is: which annotations can be combined with jsonb ones as a single API set. Johnzon has several ones but JohnzonAny has not yet a standard equivalent (we are discussing it at JSON-B) so I guess it should stay there for now and could probably be dropped once standardized.

vojtechhabarta:
Ok, seems reasonable.
If you would expect the need to support more Johnzon specific annotations we could think about having two parsers: one for generic JSON-B and second specifically for Johnzon. But I would definitely prefer to have just one if JohnzonAny is only exception.

rmannibucau:
@vojtechhabarta like a generator chain? thought about it but if you take johnzon-mapper (the pre-jsonb mapper with its own API) and accumulate it with JSON-B parser then you would have inaccurate fields (typically renamed ones would be there twice with a dead binding and filtering on the member can be random in some cases). I think it will stay the only exception and strongly hope it gets standardized so let's stay like that for now, does not cost much :). Thanks a lot for your work btw!

PR #470, opened 2020-04-02
Don't resolve deprecated compile configuration

### LINKED ISSUE #469: Deprecation warnings with Gradle 6.3
When I use the gradle plugin with Gradle 6, I get the following deprecation warning.
```
> Task :generateTypeScript
Running TypeScriptGenerator version 2.21.588
The compile configuration has been deprecated for resolution. This will fail with an error in Gradle 7.0. Please resolve the compileClasspath configuration instead. Consult the upgrading guide for further information: https://docs.gradle.org/6.3/userguide/upgrading_version_5.html#dependencies_should_no_longer_be_declared_using_the_compile_and_runtime_configurations
```

olegshtch:
Fixes #469

### DIFF
--- a/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
+++ b/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
@@ -134,13 +134,12 @@ public class GenerateTask extends DefaultTask {
         // class loader
         final Set<URL> urls = new LinkedHashSet<>();
         for (Task task : getProject().getTasks()) {
-            if (task.getName().startsWith("compile") && !task.getName().startsWith("compileTest")) {
+            if ("compileClasspath".equals(task.getName())) {
                 for (File file : task.getOutputs().getFiles()) {
                     urls.add(file.toURI().toURL());
                 }
             }
         }
-        urls.addAll(getFilesFromConfiguration("compile"));
         urls.addAll(getFilesFromConfiguration("compileClasspath"));
 
         try (URLClassLoader classLoader = Settings.createClassLoader(getProject().getName(), urls.toArray(new URL[0]), Thread.currentThread().getContextClassLoader())) {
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Thanks for this PR.
I think we will need to do this change but unfortunately it breaks some builds (for example `sample-gradle` module). So I would do it when Gradle 7 is released (at least RC version).

vojtechhabarta:
I did needed change in fd02a2d23c61b08e60e646f5827235e744faed12. I also updated `sample-gradle` module.

typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
```diff
@@ -134,13 +134,12 @@ public void generate() throws Exception {
         // class loader
         final Set<URL> urls = new LinkedHashSet<>();
         for (Task task : getProject().getTasks()) {
-            if (task.getName().startsWith("compile") && !task.getName().startsWith("compileTest")) {
+            if ("compileClasspath".equals(task.getName())) {
```
vojtechhabarta, line 137: This need to stay because otherwise classpath would not contain compiled classes.

typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
```diff
@@ -134,13 +134,12 @@ public void generate() throws Exception {
...
                 for (File file : task.getOutputs().getFiles()) {
                     urls.add(file.toURI().toURL());
                 }
             }
         }
-        urls.addAll(getFilesFromConfiguration("compile"));
         urls.addAll(getFilesFromConfiguration("compileClasspath"));
```
vojtechhabarta, line 143: 👍

typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/TypeScriptGeneratorPlugin.java
```diff
@@ -14,7 +14,7 @@ public void apply(Project project) {
         final Task generateTsTask = project.task(Collections.singletonMap(Task.TASK_TYPE, GenerateTask.class), "generateTypeScript");
 
         for (Task task : project.getTasks()) {
-            if (task.getName().startsWith("compile") && !task.getName().startsWith("compileTest")) {
+            if ("compileClasspath".equals(task.getName())) {
```
vojtechhabarta, line 17: This needs to stay because otherwise `generateTypeScript` task would not depend on compilation tasks like `compileJava` or `compileKotlin` and could use stale classes.

PR #476, opened 2020-04-16
enums must be processed, even when comming from a dependency

rmannibucau:
Issue appears when having A model extending B and B comming from a dependency. If B has some enum then it is not processed. I did the fix for JSON-B but didn't check it affects other libraries.
This is my last blocker to disable manual updates of the model :).


vojtechhabarta:
It is not only about enums, it's about all properties types. For example following test would fail:
``` java
public static class OuterObject {
    public InnerObject innerObject;
}

public static class InnerObject {
    public String text;
}

public void testPropertyTypeProcessing() {
    final String output = generate(settings, OuterObject.class);
    Assert.assertTrue(output, output.contains("interface InnerObject"));
}
```

I think other parsers like `Jackson2Parser` do this correctly.

Do you want to improve this PR (should not be difficult) or should I merge it as it is now?

rmannibucau:
@vojtechhabarta good catch, updated the pr (just dropped the num filtering)

vojtechhabarta:
Released in version [2.23.603](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.23.603).

Issue #330, opened 2019-04-10
Support "whitelistPackages" for faster scanning


antch:
As an experiment I used `ClassGraph#whitelistPackages()` to limit the packages that should be scanned for my project. This cut the runtime of typescript-generator from 22 seconds down to 15. It would be extremely helpful for this configuration option to be supported.

antch:
I have this enhancement ready to go if you're on board with it.

vojtechhabarta:
I am reluctant to add this feature for following reasons:
- users could configure whitelist inconsistently with `classPatterns` parameter which could be hard to debug
- each configuration parameter creates clutter and makes the usage less straightforward

On the other hand I agree that 15 vs. 22 seconds is significant improvement.

antch:
IMO any conflict with `classPatterns` would be pretty obvious. Where I see potential confusion is that explicitly-specified "classes" would not be subject to the whitelist (since that does not use the `Classgraph` object). My PR does specifically document that limitation FWIW.

Is there any other way aside from adding an explicit parameter, perhaps some mechanism to allow arbitrary customization or factory for the `Classgraph` instance? Though, now that I type it out that seems like a nightmare.

I am looking to break my usage down into three different executions of the generator, one for each superclass, and without the whitelisting the runtime would make it a non-starter...

antch:
I have a cleaned-up branch ready to go, definitely willing to work with you on any changes if you would consider the feature!

vojtechhabarta:
I checked some build server logs of 10 different services. Numbers of scanned classes were from 22k to 36k and scanning times were from 0.5 to 2.2 seconds. On my local machine 36k classes were scanned in 4.2 seconds which is slower but still it's not something which would require optimization.

So I am curious why it is so slow in your case. How many classes are scanned in your run? Or maybe there is some difference in caching, in my case typescript-generator is run as part of maven build so needed files may already be in file system cache.

What do you think? This would be really ideal if we could have fast runs without requiring any additional parameter.

antch:
My project is undergoing a Maven restructuring, so I wanted to wait until I could test this again as I realized I wasn't even scanning all of the classes that I needed. The difference is still ~7 seconds.

With whitelisting:
```
Scanning finished in 11.12 seconds. Total number of classes: 189788.
```

Without whitelisting:
```
Scanning finished in 3.91 seconds. Total number of classes: 28778.
```

vojtechhabarta:
[...]

PR #477, opened 2020-04-16
enable to configure maven scanning to avoid to scan the whole classpath

rmannibucau:


vojtechhabarta:
I think it is not a good idea to allow excluding some artifacts from class path. This could lead to some errors due to inconsistent dependencies.

I would like to know what's the reason behind this feature, maybe it could be solved differently.

BTW typescript-generator-maven-plugin and typescript-generator-gradle-plugin should offer the same features.

rmannibucau:
Idea is tobe able to exclude scanning of not needed dependencies, in my case 60 jars for 2 needed. It speeds up generation and will enable to reduce io/cpu usage which is way too high for what is generated on big projects.

vojtechhabarta:
I thing instead of excluding artifacts from class path it would be safer to just limit which packages are scanned. ClassGraph library (which is used for class path scanning) has `whitelistPackages` method which allows to specify this. This feature was discussed in #330.

BTW what are your times without any optimisation?

rmannibucau:
@vojtechhabarta between 4s (100% cpu) to 15 depending im on a ssd or plain old drive. Some deps are hugggeeee jars.
Package whitelisting will be slower cause you will go through the classloader.getResources if you optimize it and then it back this o(n) issue whereas this jar filtering makes it minimal. Also it will potentially miss zip index and if it is a big jar it will hurt.
In terms of safety i think it is very rare it is unsafe and the plugin could, worse case, fail in those case saying it misses a dep.

vojtechhabarta:
I don't plan to add possibility to alter classpath for speed up purposes because I consider it dangerous. Sometimes people have issues with wrong classpath and I don't want to make this even worse.

But in version [2.29.814](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.29.814) I added [scanningAcceptedPackages](https://www.habarta.cz/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html#scanningAcceptedPackages) parameter which allows to speed up the classpath scanning by specifying allowed packages.

PR #478, opened 2020-04-22
gradle-plugin: don't propagate gradle dependency

bennofs:
Changing the scope to "provided" makes sure that the gradle api
dependency does not leak into the consumers of the plugin.

The issue came up in a gradle project: since typescript-generator has a dependency on Gradle API 3, this dependency leaked into the runtime classpath of my build script which caused issues. 

Since the plugin can only be used in a gradle build, the gradle api should already be part of the runtime classpath and doesn't need to propagated.


vojtechhabarta:
Thanks for your PR. Merged.

vojtechhabarta:
Released in version [2.23.603](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.23.603).

PR #483, opened 2020-05-21
ensure JSON-B parser makes attributes optional when not final

rmannibucau:
Small enhancement to enable JSON-B parser to also handle optional attributes.

This PR assumes final fields means not optional (until settings are configured explicitly). This is not 100% true in java (except for primitives) but I assume it is a good compromise overall.

Let me know if you think we should make all objects optional.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/JsonbParser.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/JsonbParser.java
@@ -1,6 +1,7 @@
 package cz.habarta.typescript.generator.parser;
 
 import cz.habarta.typescript.generator.ExcludingTypeProcessor;
+import cz.habarta.typescript.generator.OptionalProperties;
 import cz.habarta.typescript.generator.Settings;
 import cz.habarta.typescript.generator.TypeProcessor;
 import cz.habarta.typescript.generator.util.PropertyMember;
@@ -9,15 +10,18 @@ import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Proxy;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
@@ -34,6 +38,7 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.json.bind.annotation.JsonbCreator;
 import javax.json.bind.annotation.JsonbProperty;
 import javax.json.bind.annotation.JsonbTransient;
 import javax.json.bind.annotation.JsonbVisibility;
@@ -137,6 +142,36 @@ public class JsonbParser extends ModelParser {
         }
 
         private List<PropertyModel> visit(final Class<?> clazz) {
+            return Stream.of(clazz.getConstructors())
+                    .filter(it -> it.isAnnotationPresent(JsonbCreator.class))
+                    .findFirst()
+                    .map(it -> new ArrayList<>(Stream.concat(visitConstructor(it), visitClass(clazz).stream())
+                            .collect(Collectors.toMap(PropertyModel::getName, Function.identity(), (a, b) -> a)) // merge models
+                            .values()))
+                    .orElseGet(() -> new ArrayList<>(visitClass(clazz)));
+        }
+
+        private Stream<PropertyModel> visitConstructor(final Constructor<?> constructor) {
+            // JSON-B 1.0 assumes all constructor params are required even if impls can diverge on that due
+            // to user feedbacks so for our libraryDefinition let's assume it is true.
+            // only exception is about optional wrappers which can be optional indeed
+            return Stream.of(constructor.getParameters())
+                    .map(it -> {
+                        final Type type = it.getParameterizedType();
+                        return new PropertyModel(
+                                Optional.ofNullable(it.getAnnotation(JsonbProperty.class))
+                                    .map(JsonbProperty::value)
+                                    .filter(p -> !p.isEmpty())
+                                    .orElseGet(it::getName),
+                                type,
+                                isOptional(type) || OptionalInt.class == type ||
+                                        OptionalLong.class == type || OptionalDouble.class == type,
+                                new ParameterMember(it),
+                                null, null, null);
+                    });
+        }
+
+        private List<PropertyModel> visitClass(final Class<?> clazz) {
             return delegate.find(clazz).entrySet().stream()
                     .filter(e -> !isTransient(e.getValue(), visibility))
                     .filter(e -> johnzonAny == null || e.getValue().getAnnotation(johnzonAny) == null)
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Thanks for another PR.

Regarding optional properties first thing is that typescript-generator has 3 modes depending on `optionalProperties` parameter. Value `all` means that all properties are marked as optional. Value `useSpecifiedAnnotations` means that optionality is indicated by annotations specified using `optionalAnnotations` or `requiredAnnotations` parameter. Third value `useLibraryDefinition` means that it depends on library (in this case JSON-B) how it defines what is optional. It would be good if `JsonbParser` supported these modes.

To implement first two modes it is just matter of calling `isPropertyOptional` method (from `ModelParser` class) but it needs wrapped member on input. Here are some example tests:
``` java
    @Test
    public void testOptionalProperties() {
        settings.optionalProperties = OptionalProperties.all;
        final String output = generate(settings, ObjecToTestOptionality.class);
        Assert.assertTrue(output, output.contains(" foo?:"));
        Assert.assertTrue(output, output.contains(" bar?:"));
    }

    @Test
    public void testOptionalPropertiesMarkedByAnnotation() {
        settings.optionalProperties = OptionalProperties.useSpecifiedAnnotations;
        settings.optionalAnnotations.add(Nullable.class);
        final String output = generate(settings, ObjecToTestOptionality.class);
        Assert.assertTrue(output, output.contains(" foo:"));
        Assert.assertTrue(output, output.contains(" bar?:"));
    }

    public static class ObjecToTestOptionality {
        public String foo;
        @Nullable
        public String bar;
    }
```

For third mode (`useLibraryDefinition`) I am not sure. For example Jackson library has `@JsonProperty(required = false)` annotation. But it seems to me that JSON-B doesn't have such thing. And it also seems to me that `final` modifier doesn't have any impact. When client is sending JSON which is then read by JSON-B all properties are treated as optional (regardless `final` modifier). But when JSON-B writes JSON which is then consumed by client it would not be much convenient to test all properties whether they are `undefined` (at least in TypeScript strict mode). What do you thing?

rmannibucau:
Hi @vojtechhabarta , isnt generator handling done with https://github.com/vojtechhabarta/typescript-generator/pull/483/files#diff-a916707a998b8004c8c9fe4342264d56R169 (expected it) ?

Regarding the library specific implementation there is nothing really in JSON-B (JsonbNillable is more about JSON null than model requirements which are two transversal things) but statiscally when you put final you set the value in the constructor and today JSON-B tends to requires a value (even we if discuss to relax it) whereas in all other case it is optional.

Now the goal of this PR is really to enable to make them all optional by default since it is what this kind of model is generally, having them required by default does not reflect most of applications AFAIK - I'm not discussing if it is good or bad ;).

vojtechhabarta:
Hi, I overlooked that you are adding this call (`isPropertyOptional`) in the PR and I only tried mentioned tests on current code. Sorry for that.

There is one thing that should be changed: library specific logic should not be combined with `isPropertyOptional` call (using `||`). Only one of them should be used at the same time. When `optionalProperties` parameter is set to `useLibraryDefinition` specific logic should be used. In other cases `isPropertyOptional` should be called. See https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java#L228. Following example illustrates the reason:
``` java
@Test
public void testRequiredPropertyMarkedByAnnotation() {
    settings.optionalProperties = OptionalProperties.useSpecifiedAnnotations;
    settings.requiredAnnotations.add(Required.class);
    final String output = generate(settings, ObjecWithRequiredProperty.class);
    Assert.assertTrue(output, output.contains(" foo"));
    Assert.assertTrue(output, output.contains(" bar?"));
}
public static class ObjecWithRequiredProperty {
    @Required
    public String foo;
    public String bar;
}
@Retention(RetentionPolicy.RUNTIME)
public @interface Required {
}
```
Here `foo` property should be required regardless `final` modifier or other logic.

Regarding `final` modifier: let's use following example just to be sure:
``` java
public static class ObjecToTestOptionality2 {
    public final String foo;
    public final String bar = "SOME_CONSTANT";  // not much common

    @JsonbCreator
    public ObjecToTestOptionality2(String foo) {
        this.foo = foo;
    }
}
```
There is constructor annotated with `@JsonbCreator` so the `foo` property is required. And you are also saying that usually when I have `final` field I initialize it in such constructor. I agree that in most cases this is true. So using `final` modifier for determining optionality is probably good approximation. But maybe it would not be so difficult to use parameters of creator constructor which I think is the source of truth.

rmannibucau:
@vojtechhabarta updated the PR, you are perfectly right the serialization model is more about the constructor in such cases than the fields wich can reflect something totally unrelated.

vojtechhabarta:
Great.
I still see one thing: if I add creator constructor to `ObjecWithRequiredProperty` test class:
``` java
public static class ObjecWithRequiredProperty {
    @RequiredAnnotation
    public String foo;
    public String bar;
    @JsonbCreator
    public ObjecWithRequiredProperty(String foo, String bar) {}
}
```
then `testRequiredPropertyMarkedByAnnotation` fails even is has `useSpecifiedAnnotations`. Ceator constructor should not have any impact in this case.

rmannibucau:
@vojtechhabarta this case is more complex, let me detail what is happening:

1. constructor is selected and both parameters are taken as model properties with their configuration (name, optional=false, required, ...)
2. fields are taken as properties with their model
3. both sets are merged ignoring fields already existing in constructor

So overall it means @RequiredAnnotation should go into the constructor there until we merge both annotation set (which is not done in JSON-B since there is no way to know foo field matches foo constructor parameter).

So I tend to think this is not something we want to support in this version.

Wdyt?

vojtechhabarta:
Ok, understood.

In https://github.com/vojtechhabarta/typescript-generator/commit/071c9466085bee883cccaf2bfe3a6343f5bafe9e I added method for reading types of constructor parameters. You can use it in `visitConstructor` method for example like this:
``` java
final List<Type> parameterTypes = settings.getTypeParser().getConstructorParameterTypes(constructor);
final List<Pair<Parameter, Type>> parameters = Utils.zip(Arrays.asList(constructor.getParameters()), parameterTypes);
return parameters.stream()
        .map(pair -> {
            final Parameter it = pair.getValue1();
            final Type type = pair.getValue2();
            final Optional<JsonbProperty> property = Optional.ofNullable(it.getAnnotation(JsonbProperty.class));
            final PropertyMember propertyMember = new PropertyMember(it, pair.getValue2(), it.getAnnotatedType(), it::getAnnotation);
```
(`PropertyMember` is needed as parameter to `isPropertyOptional` method.)
TypeParser is used for handling of nullable types which allows for example: `public List<@Nullable String> baz` mapped to `baz?: (string | null)[]`.

But this is something which can be improved sometime in the future.

I just noticed that you have your own implementation of `GenericArrayType`. There was already implementation in `cz.habarta.typescript.generator.type.JGenericArrayType`. Could you please check if it could be used?

rmannibucau:
@vojtechhabarta updated

vojtechhabarta:
@rmannibucau thank you very much for your work

vojtechhabarta:
@rmannibucau if you had time it would be cool if you could also implement nullable types. This means that for example for Java property `public List<@Nullable String> foos` generated TypeScript property can be `foos: (string | null)[]`. Here nullable annotation is for example `org.checkerframework.checker.nullness.qual.Nullable` (must have `TYPE_USE` target).

Here are example tests:
``` java
import org.checkerframework.checker.nullness.qual.Nullable;

public static class ListOfNullableElementsField {
    public List<@Nullable String> foos;
}

public static class ListOfNullableElementsGetter {
    public List<@Nullable String> getFoos() {
        return null;
    }
}

public static class ListOfNullableElementsConstructor {
    private List<String> foos;

    @JsonbCreator
    public ListOfNullableElementsConstructor(List<@Nullable String> foos) {
        this.foos = foos;
    }
}

@Test
public void testNullabilityField() {
    settings.nullableAnnotations.add(Nullable.class);
    final String output = generate(settings, ListOfNullableElementsField.class);
    System.out.println(output);
    Assert.assertTrue(output, output.contains(" foos: (string | null)[];"));
}

@Test
public void testNullabilityGetter() {
    settings.nullableAnnotations.add(Nullable.class);
    final String output = generate(settings, ListOfNullableElementsGetter.class);
    System.out.println(output);
    Assert.assertTrue(output, output.contains(" foos: (string | null)[];"));
}

@Test
public void testNullabilityConstructor() {
    settings.nullableAnnotations.add(Nullable.class);
    final String output = generate(settings, ListOfNullableElementsConstructor.class);
    System.out.println(output);
    Assert.assertTrue(output, output.contains(" foos: (string | null)[];"));
}
```

Test of nullable type used in creator constructor is already passing. For remaining two it is needed to use `getFieldType()` and `getMethodReturnType()` methods from `settings.getTypeParser()`. Those methods also take care of Kotlin's nullable types like `List<String?>`.

vojtechhabarta:

PR #487, opened 2020-06-01
[jsonb] use processTypeAndCreateProperty to miss less types to process

rmannibucau:
side note: think other parsers (gson?) have this bug too


vojtechhabarta:
Thanks for this PR.
You are right, GsonParser should use this method too.

vojtechhabarta:

Issue #489, opened 2020-06-03
Do not create constructors for Typescript interfaces


manstis:
If `generateConstructors` is set to `true` Java interfaces result in Typescript interfaces with a constructor. This is invalid. 

As an aside, to also explain my second commit on the PR, if using `generateConstructors` and the `JsonDeserializationExtension` together, the extension attempts to call zero-parameter constructors. I've updated the extension to pass the `data` parameter passed to the JSON related helper method to the object constructor; matching that of the generated constructor.

manstis:
See https://github.com/vojtechhabarta/typescript-generator/pull/490 for a fix.

vojtechhabarta:
Released in [2.24.612](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.24.612).

PR #490, opened 2020-06-03
ISSUE-489: Do not create constructors for Typescript interfaces

manstis:
See https://github.com/vojtechhabarta/typescript-generator/issues/489

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
@@ -519,13 +519,17 @@ public class ModelCompiler {
                     )));
                 }
             }
-            final TsConstructorModel constructor = new TsConstructorModel(
-                    TsModifierFlags.None,
-                    Arrays.asList(new TsParameterModel("data", dataType)),
-                    body,
-                    /*comments*/ null
-            );
-            beans.add(bean.withConstructor(constructor));
+            if (bean.isClass()) {
+                final TsConstructorModel constructor = new TsConstructorModel(
+                        TsModifierFlags.None,
+                        Arrays.asList(new TsParameterModel("data", dataType)),
+                        body,
+                        /*comments*/ null
+                );
+                beans.add(bean.withConstructor(constructor));
+            } else {
+                beans.add(bean);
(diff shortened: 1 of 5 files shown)


manstis:
I've also added a commit that fixes `JsonDeserializationExtension` when the `generateConstructors` flag has been set to `true` (in which cases there is no zero-parameter constructor generated and we need to pass the `data` parameter). 

I had to expose `Settings` to `ModelTransformer` - that I thought reasonable - IDK how you'd want to tackle backwards compatibility? I could keep the original methods on `ModelTransformer`, mark as deprecated and add new ones with the `Settings` parameter. The deprecated methods would call the new ones passing `null` and I'd add a `null` check before trying to ascertain whether constructors had been generated...

All good stuff.. enjoying this plugin so far :-)

vojtechhabarta:
First commit (fixing constructors in interfaces) is clear. I would merge it.

Incompatibility in second commit is problematic. In typescript-generator I don't consider Java compatibility as important as compatibility of Maven and Gradle plugins (because it has fewer users and Java compiler can help) but this part of Java API is explicitly meant for users.

However I think in this case it is not needed to pass `Settings` to `ModelTransformer` method. Solution could be to detect if there is explicit constructor using `bean.getConstructor() != null` check (instead of `settings.generateConstructors`).

manstis:
Thank-you @vojtechhabarta 

I will revise the second commit following your proposal and let you know when complete.

manstis:
Hi @vojtechhabarta I've updated the second commit. 

Please let me know if you would like further changes.

vojtechhabarta:
Hi @manstis,
thank you for your changes. Merging...

manstis:
Hi @vojtechhabarta thank-you kindly.

My $1M question is.... when are you planning your next release?

It may not surprise you that I encountered the issue whilst using your plugin for a work task...

I have no strict deadline but knowledge of when you next plan to release would be helpful.

Thank-you,

Mike

vojtechhabarta:
Hi Mike,
I usually plan release once a month. So it should be soon but I don't want to promise any date.
Vojta

manstis:
Hi @vojtechhabarta no problem. 

I have months of development work remaining so no hurry.

vojtechhabarta:

PR #495, opened 2020-06-23
@ModelAttribute in Spring

kpolszewski:
When using many Query parameters in Spring MVC it is convinient to wrap them into one object instead of multiple @RequestParam annotations

Example is available here, point 9: https://reversecoding.net/spring-mvc-requestparam-binding-request-parameters/

Currently, if we drop annotations generator skips this kind of objects - it would be great to support them.

To prevent some unexpected errors while upgrading , we could force to use @ModelAttribute annotation if we want to enable this behaviour. Said annotation is current default fallback, according to the bottom of this documentation section: https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-arguments

You can find my proposal in PR.


vojtechhabarta:
Thanks for this PR with good description.
I agree that requiring `@ModelAttribute` annotation is reasonable solution now.
To support parameters without an annotation ("Any other argument" row in [Method Arguments](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-arguments) table) we would first need to handle all rows in this table.

vojtechhabarta:

PR #497, opened 2020-06-29
Jackson2: Super class did not get type information

michaelzangl:
Jackson can (de)serialize instances of the hierarchy root class when using @JsonSubTypes. When generating interface for this java class:

```java
    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
    @JsonSubTypes(@JsonSubTypes.Type(NonAbstractRootSub.class))
    @JsonTypeName("rootType")
    public static class NonAbstractRoot {
    }

    @JsonTypeName("subType")
    public static class NonAbstractRootSub extends NonAbstractRoot {
    }
```

Previously wrong generated code ("rootType" missing):
```ts
interface NonAbstractRoot {
    type: "subType";
}

interface NonAbstractRootSub extends NonAbstractRoot {
    type: "subType";
}

type NonAbstractRootUnion = NonAbstractRootSub;
```

Correct:
```ts
interface NonAbstractRoot {
    type: "rootType" | "subType";
}

interface NonAbstractRootSub extends NonAbstractRoot {
    type: "subType";
}

type NonAbstractRootUnion = NonAbstractRootSub;
```

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
@@ -253,7 +253,7 @@ public class Jackson2Parser extends ModelParser {
         if (isSupported(jsonTypeInfo)) {
             // this is parent
             discriminantProperty = getDiscriminantPropertyName(jsonTypeInfo);
-            discriminantLiteral = null;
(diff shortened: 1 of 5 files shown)


vojtechhabarta:
Thanks for your pull request with good example.

While this fixes typescript-generator inconsistency with Jackson it can break some users code. I am not sure how frequent this problem could be but I think we need to change this sooner the better.

For users:

**make root class `abstract` if you don't want discriminant value for this class to be included in union**

vojtechhabarta:

PR #501, opened 2020-07-22
Fix disappearing Spring URL trailing slash

AMilassin:
## Issue
The URL's for the following Controller were incorrectly generated as the trailing slash is always omitted, but it's an important distinction of the URL.
```
    @RestController
    @RequestMapping("/controller")
    public class TestUrlSuffixController {
        @GetMapping("/")
        public void response() {
        }
        @GetMapping("")
        public void response2() {
        }
    }
```

The generated URL controller for both mappings were `controller` instead of `controller/` and `controller` respectively. Having or not having a trailing slash also means a different URL.

## Solution
Rewrite of `Utils.joinPath()`.
Added tests for both the initial issue and the `joinPath` rewrite.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/util/Utils.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/util/Utils.java
@@ -50,24 +50,37 @@ public final class Utils {
     }
 
     public static String joinPath(String part1, String part2) {
-        final List<String> parts = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


vojtechhabarta:
Thanks for investigating the issue, fixing it and writing tests.
I just changed the `joinPath` method to be more readable (at least to me) and added two more tests.
Thanks again.

vojtechhabarta:

PR #532, opened 2020-08-12
Support for Vavr collections and Option type

nurkiewicz:
[Vavr](https://www.vavr.io/) is a functional programming library for Java. Unlike Guava, Vavr collections do not inherit from standard Java collection API. Therefor `typescript-generator` does not recognize them. This PR adds support for Vavr types so they are treated as collections or optional types.

### DIFF
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -94,6 +94,11 @@
             <artifactId>js-scriptengine</artifactId>
             <version>${graalvm.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.vavr</groupId>
+            <artifactId>vavr</artifactId>
+            <version>0.10.3</version>
+        </dependency>
         <!--test dependencies-->
         <dependency>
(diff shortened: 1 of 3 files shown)


nurkiewicz:
Hi, any news on this PR?

vojtechhabarta:
Sorry for very late response.

Thank you for your PR and interest in adding support for Vavr library. I don't want to add dependencies on different libraries to typescript-generator so in #545 I implemented general mechanism how to add support for data libraries. I would be very glad if you could try it and check if it does what you need (it is in `master` branch, not yet released). You can also see the [definition](https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-core/src/main/resources/datalibrary/vavr.json) of Vavr library.

vojtechhabarta:
#545 released in [v2.26.723](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.26.723).

nurkiewicz:
Thank you @vojtechhabarta, your solution seems way more robust!

PR #533, opened 2020-08-14
add skip option for a more modular configuration

sadraskol:
Hello !

`skip` parameter is supported by core maven plugins. It allows to easily configure large multi-module project using the parameter inheritance rather than the complex plugins configured by profile inheritance.

As the PR shows, the implementation is pretty straight forward and it would be appreciated by some of maven users.

I am not sure the `TypeScriptGenerator#generateTypeScript` I considered the entrypoint is the right one. Maybe you can see a better place to do it.

Thanks!

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -128,6 +128,7 @@ public class Settings {
     public boolean jackson2ModuleDiscovery = false;
     public List<Class<? extends Module>> jackson2Modules = new ArrayList<>();
     public ClassLoader classLoader = null;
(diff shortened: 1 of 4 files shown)


vojtechhabarta:
Hello,
thanks for your PR and sorry for late response.

This feature is Maven specific. I realized that Gradle has built-in mechanisms like `-x` parameter to skip task execution. (And when typescript-generator is invoked from Java it can be controlled directly of course.) So in my ab978fa commit I added `skip` feature just in `GenerateMojo` class (not in `Settings` and `TypeScriptGenerator` classes). Otherwise you were right about entry point.

I also added `typescript.generator.skip` property for skipping typescript-generator from command line.

vojtechhabarta:

PR #541, opened 2020-09-17
Sort constructor property assignments when sortDeclarations is true

jpstratman:
Previously, `sortDeclarations = true` would cause only the generated class properties to be sorted, but the constructor assignment ordering remained unchanged. This defeats one of the main use cases of `sortDeclarations`, which is for generated code to be unchanged when the source code property ordering changes.

When `sortDeclarations = true`:
Before this change:
```typescript
export class FooBar {
	bar: number;
	foo?: string;

	constructor(data: FooBar) {
		this.foo = data.foo;
		this.bar = data.bar;
	}
}
```
After this change (sorted property assignments in constructor):
```typescript
export class FooBar {
	bar: number;
	foo?: string;

	constructor(data: FooBar) {
		this.bar = data.bar;
		this.foo = data.foo;
	}
}
```

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java
@@ -121,7 +121,7 @@ public class ModelCompiler {
         tsModel = removeInheritedProperties(symbolTable, tsModel);
         tsModel = addImplementedProperties(symbolTable, tsModel);
         if (settings.generateConstructors) {
-            tsModel = addConstructors(symbolTable, tsModel);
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Thanks for your PR.
I agree that for easy readability the assignments in constructors should be in the same order as properties.
Merged.

Just one note: line 514 (https://github.com/vojtechhabarta/typescript-generator/pull/541/files#diff-8e6a80fd6c309e32cff93310315f1092R514) could throw `NullPointerException` because `sortDeclarations` parameter was `Boolean` and could be `null`. But instead of changing it to `boolean` I removed it completely because this method can use `settings.sortDeclarations` directly.

Thanks again.

vojtechhabarta:

PR #545, opened 2020-09-28
Support for data libraries (guava, joda, vavr)

vojtechhabarta:
This is implementation of general mechanism for mapping certain Java types from supported library to TypeScript without modifying typescript-generator Java source code and adding dependency on supported library.

Let's first see how these libraries are supported in Jackson2. Jackson2 has concept of modules so for example for [guava](https://github.com/google/guava) library there is a [GuavaModule](https://github.com/FasterXML/jackson-datatypes-collections/blob/master/guava/src/main/java/com/fasterxml/jackson/datatype/guava/GuavaModule.java) (in `com.fasterxml.jackson.datatype:jackson-datatype-guava`) that adds support for guava to Jackson2 by adding serializers and deserializers for certain guava data types.

To map those types correctly in typescript-generator previously it was needed to use `customTypeMappings` parameter to add mappings for needed types manually. Instead of this it is now possible to add `guava` mappings using `additionalDataLibraries` parameter. Here is Maven example:

``` xml
<configuration>
    <additionalDataLibraries>
        <library>guava</library>
    </additionalDataLibraries>
    ...
</configuration>
```

Supported libraries are:
- `guava` ([source](https://github.com/google/guava)) ([definition](https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-core/src/main/resources/datalibrary/guava.json))
- `joda` ([source](https://github.com/JodaOrg/joda-time)) ([definition](https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-core/src/main/resources/datalibrary/joda.json))
- `vavr` ([source](https://github.com/vavr-io/vavr)) ([definition](https://github.com/vojtechhabarta/typescript-generator/blob/master/typescript-generator-core/src/main/resources/datalibrary/vavr.json))

Note: typescript-generator doesn't have dependency on those libraries (except for tests) but it needs to load classes from libraries definitions to be able to also recognize their subclasses. So it relies on user project dependencies.

As mentioned this feature processes also subclasses which is different from `customTypeMappings` parameter that can't map subclasses (yet).

This PR also changes how built-in (known) types are handled. Previously there were individual mappings (without subclasses) whereas now there are mappings for supertypes like `Number` (which matches also for example `BigDecimal` and `com.google.common.primitives.UnsignedInteger`) or `Date` (which matches also for example subclasses in `java.sql` package). And also mapping for `Calendar` class was added.

This PR is related to #532 (Vavr support). Difference is that this PR doesn't require adding Vavr dependency. It also adds mapping for more Vavr data types.

This PR is also related to #435 (custom mappings of supertypes). It prepares some logic (subclasses, generics) for easier implementation.

### DIFF
--- a/typescript-generator-core/pom.xml
+++ b/typescript-generator-core/pom.xml
@@ -19,6 +19,7 @@
         <jackson2.version>2.11.1</jackson2.version>
         <jersey.version>2.31</jersey.version>
         <graalvm.version>20.1.0</graalvm.version>
+        <vavr.version>0.10.3</vavr.version>
     </properties>
 
     <dependencies>
@@ -120,6 +121,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+            <version>${jackson2.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.glassfish.jersey.containers</groupId>
             <artifactId>jersey-container-jdk-http</artifactId>
             <version>${jersey.version}</version>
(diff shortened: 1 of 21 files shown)


vojtechhabarta:

PR #552, opened 2020-09-30
Javadoc generation : use class field's javadoc as a default when no javadoc is found on getter/setter

cbrethes:
just lovin this plugin <3

I use it to generate typescript and pasted like a monkey the (very good and working) tutorial to generate Javadoc from one project to another and works like a charm but i'm just facing one problem : 
I use lombok getter/setter generation so i write Javadoc directly on class attributes.
The Javadoc is well generated (just added a `<show>private</show>` in maven-javadoc-plugin configuration) when i inspect javadoc.xml ait is not inserted in typescript.

I think it would be nice to search for javadoc on getters/setters if any but to fallback to property's javadoc

This is the content of this pull request

Thx for your work !


vojtechhabarta:
Thanks for your kind words.
Merged.

vojtechhabarta:
Released in [v2.27.744](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.27.744).

PR #587, opened 2020-12-03
Added setting for formatting enum keys in pascal case

jrparish:
Our API team likes to use UNDERSCORE_CASE enums, but UI team prefers the more readable pascal casing. 

This adds an optional setting to format the keys accordingly.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -83,6 +83,7 @@ public class Settings {
     private List<CustomTypeAlias> validatedCustomTypeAliases = null;
     public DateMapping mapDate; // default is DateMapping.asDate
     public EnumMapping mapEnum; // default is EnumMapping.asUnion
+    public boolean pascalCaseEnums = false;
     public boolean nonConstEnums = false;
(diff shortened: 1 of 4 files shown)


vojtechhabarta:
Thanks for this feature. I think it is nice so I am merging it but I will push few changes.

Firstly I am adding `enumMemberCasing` parameter (instead of special parameter for Pascal case). This parameter can support not only "PascalCase" but also "camelCase" and possibly others. Since configuration parameters should be usable from build plugins (Maven, Gradle) I am adding this new parameter there as well. There is also description which is used when generating [documentation](https://www.habarta.cz/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html#enumMemberCasing). Here is an example of usage from Maven:

``` xml
<configuration>
    ...
    <mapEnum>asEnum</mapEnum>
    <enumMemberCasing>PascalCase</enumMemberCasing>
    ...
</configuration>
```

Next are some improvements in splitting identifier into words using regular expression. It should now support letters other than `A-Z` and numbers should be treated as words. I am still not sure it supports all cases well but at least I added unit test specifically for this part so we can easily test and improve other cases.

Could you please check my changes if the feature works as expected?

jrparish:
That looks great, thank you!

vojtechhabarta:
Released in [v2.28.785](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.28.785).

Issue #600, opened 2020-12-17
Non-deterministic output for Json Deserialization plugin despite sortDeclaration property enabled


kevincmanning:
Getting a different order every time I run the build.

Here is a PR that defaults the initializers to sorted

kevincmanning:
https://github.com/vojtechhabarta/typescript-generator/pull/601

vojtechhabarta:
You are right that `JsonDeserializationExtension` doesn't respect `sortDeclarations` configuration parameter. And currently it's not possible to implement it because extensions don't have access to `Settings`.

But I think the goal for `JsonDeserializationExtension` should be to generate assignments in the same order as properties are declared. This order is sometimes non-deterministic (for example when defining properties using getters) and sometimes deterministic (for example when defining properties using fields).

So I moved sorting of bean properties to an earlier stage in the processing. This ensures that subsequent steps have properties in right order and can generate their declarations in correct order. This applies not only to `JsonDeserializationExtension` (and possibly to other extensions) but also to assignments in constructors (if `generateConstructors` parameter is used).

Could you please try it?

vojtechhabarta:
Released in [v2.28.785](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.28.785).

kevincmanning:
Thanks! I will try it tomorrow when I am back at work after the holidays.

PR #601, opened 2020-12-17
Adding deterministic outcome for deserialization plugin initializer method fromData

kevincmanning:
Adding deterministic outcome for deserialization plugin initializer method fromData


mabe02:
:+1:

drex44:
👍

vojtechhabarta:
I think `JsonDeserializationExtension` should not always sort assignments. Ideally it should respect `sortDeclarations` parameter which is not currently possible. So in 1c6b47df2201d8fbfa90b35ec31db3f3c6f1842a commit I solved this differently, see my comment here https://github.com/vojtechhabarta/typescript-generator/issues/600#issuecomment-752391044.

Issue #599, opened 2020-12-17
After update to 2.27 fails to generate due to constructor


mduft:
I have updated the generator
```
-        classpath 'cz.habarta.typescript-generator:typescript-generator-gradle-plugin:2.26.723'
+        classpath 'cz.habarta.typescript-generator:typescript-generator-gradle-plugin:2.27.744'
```
Now I'm receiving an error when generating DTOs:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':ui:genDtos'.
> Unexpected member 'java.lang.reflect.Constructor' in property 'sorter' in class 'io.bdeploy.interfaces.plugin.PluginInfoDto'
```
The `PluginInfoDto` has a member of type `CustomProductVersionSorter` named `sorter` - which is the property causing the failure. This field is marked with `@JsonIgnore` like so:
```
public class PluginInfoDto {
...
    @JsonIgnore
    public CustomProductVersionSorter sorter;
...
}
```
This works with the older version, so for now I'm sticking with it :)

kevincmanning:
I also got this error, and have committed some changes that fixes it locally for me. In my case, it was caused by Jackson having inconsistent names for the serializable properties and the deserializable properties.

@vojtechhabarta Please see my PR and adjust accordingly as needed or advise on best way to fix this

https://github.com/vojtechhabarta/typescript-generator/pull/602

I added a //TODO because it seems that in the event that the code is designed to add any deserializable properties to the list if they don't exist in the serializable properties list. It seems in my case it was yielding CreatorProperty which was yielding Constructor rather than Field or Method. My instinct is if that is the case, just ignore it and keep processing, but please advise.

vojtechhabarta:
@mduft, @kevincmanning thanks for reporting this problem and proposing fix.

In b2e11b1eaa2817b6fe794b230691e056fc538c52 I implemented support for `@JsonCreator` used on both constructors and factory methods (see tests).
At the same time I changed handling of unsupported cases so instead of throwing an exception it is just logged.
Could you please try it?

vojtechhabarta:
Released in [v2.28.785](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.28.785).

mduft:
Hey @vojtechhabarta - I still was on holydays :) I now tried it and it seems to work like a charm, thanks!

vojtechhabarta:
Thanks

PR #602, opened 2020-12-17
Fixing exceptions with properties that were named slightly differentl…

kevincmanning:
…y in the List of serializedProperties vs deserializedProperties


drex44:
👍

vojtechhabarta:
I implemented support for `@JsonCreator` used on both constructors and factory methods, see my comment https://github.com/vojtechhabarta/typescript-generator/issues/599#issuecomment-751666263.

Issue #603, opened 2020-12-24
Transform @Deprecated annotations into comments


SButterfly:
Hi
I marked some properties in POJO classes as `@Deprecated(since = "don't use it")`, hoping frontend team will try not to use it.
But unfortunately this annotations are not passed to typescript.
Can they be passed smth like this:
```java
public class Person {
    public String fullName;
    @Deprecated(since = "Use fullname")
    public String name;
}
```

as

```ts
interface Person {
    fullName: string;
    /** @deprecated since: Use fullname */
    name: string;
}
```

vojtechhabarta:
I implemented support for `@Deprecated` annotation used on classes, properties, enums, enum members and REST methods.

Element `since` is not supported because it was added in Java 9 but typescript-generator still supports Java 8. By the way, value of this element should be a version, not a reason.

vojtechhabarta:
Released in [v2.28.785](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.28.785).

PR #610, opened 2021-01-03
Support @Deprecated #since & #forRemoval in comment's generation

SButterfly:
Hi!
I supported methods `since` and `forRemoval` in `@Deprecated` annotation, which are only introduced in java9.
#603

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/DeprecationEnricher.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/DeprecationEnricher.java
@@ -5,7 +5,10 @@ import cz.habarta.typescript.generator.compiler.EnumMemberModel;
 import cz.habarta.typescript.generator.util.Utils;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
@@ -14,6 +17,8 @@ import java.util.stream.Collectors;
 
 public class DeprecationEnricher {
 
+    public static final String DEPRECATED = "@deprecated";
+
     public Model enrichModel(Model model) {
         final List<BeanModel> beans = mapList(model.getBeans(), this::enrichBean);
         final List<EnumModel> enums = mapList(model.getEnums(), this::enrichEnum);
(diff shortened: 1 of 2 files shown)


typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/DeprecationEnricher.java
```diff
@@ -14,21 +17,23 @@
...
+
     public Model enrichModel(Model model) {
         final List<BeanModel> beans = mapList(model.getBeans(), this::enrichBean);
         final List<EnumModel> enums = mapList(model.getEnums(), this::enrichEnum);
         final List<RestApplicationModel> restApplications = mapList(model.getRestApplications(), this::enrichRestApplication);
         return new Model(beans, enums, restApplications);
     }
 
-    private BeanModel enrichBean(BeanModel bean) {
+    public BeanModel enrichBean(BeanModel bean) {
```
SButterfly, line 29: Made most methods public to enable direct call in custom plugin.

vojtechhabarta: These methods are internal and we should keep possibility to refactor this class and remove these methods or change their signature. Could you call just `enrichModel` method from your plugin? It should not duplicate `@deprecated` tags.

In fact I considered whole class internal and not intended it to be called from other places but I understand that it could be useful.

typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/DeprecationEnricher.java
```diff
@@ -40,46 +45,90 @@ private PropertyModel enrichProperty(PropertyModel property) {
...
+            .filter(m -> m.getName().equals("since"))
+            .map(m -> {
+                try {
+                    return (String) m.invoke(deprecated);
+                } catch (IllegalAccessException | InvocationTargetException e) {
+                    return "";
+                }
+            })
+            .findFirst().orElse("");
     }
```
vojtechhabarta, line 115: For getting annotation elements using reflections there is a method [getAnnotationElementValue](https://github.com/vojtechhabarta/typescript-generator/blob/95a9bf5fc16056c507dcf330584b46a5b0be2bf1/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/util/Utils.java#L194).

SButterfly:
Moved convertToComment to utils class for reusing. And other code review fixes

vojtechhabarta:
Thanks for your PR. Merged. I will just adjust generated text.

vojtechhabarta:
Released in [2.29.814](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.29.814).

PR #619, opened 2021-01-13
allow configuration of typescriptVersion in the maven plugin

agaengel:
allow configuration of typescriptVersion in the maven plugin


vojtechhabarta:
Thanks for the PR. Merging but I will rename the parameter to `npmTypescriptVersion` (to clearly communicate the purpose of this parameter). I will also add it to Gradle plugin (it should have the same set of parameters as Maven plugin).

vojtechhabarta:
Released in [2.29.814](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.29.814).

PR #630, opened 2021-02-04
Fixing flaky tests

Shafiuzzaman-Hira:
Nondex tool identifies 59 flaky tests in this project. You can find the full list here: http://mir.cs.illinois.edu/flakytests/unfixed.html. 
I see all of the flaky tests can be fixed with a single fix. Here, I am encoding the output back into JSON using a mapper and then writing it back out again but sorted. This removes the test flakiness caused by the order of the json components. This pull request fix a single test. If you accept this, I may fix all of those 59 tests.

### DIFF
--- a/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/ReadOnlyWriteOnlyTest.java
+++ b/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/ReadOnlyWriteOnlyTest.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.google.gson.Gson;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
@Shafiuzzaman-Hira I think that when testing some business logic it is good to deal with different order of JSON properties. But here we have different case. The purpose of tests is mainly to verify what exactly is generated. So sometimes the order is not important but sometime it is. By the way, in this particular test you are not parsing JSON but TypeScript code.

Thank you for your offer but I would keep the tests as they are now.

Shafiuzzaman-Hira:
I got your point. However, the implementation relies on the non-deterministic method java.lang.Class#getDeclaredFields() which can change the order. Consider the case of upgrading from Java 6 to Java 7 that changed the order in which the Java Reflection API returned the list of methods in a class. JUnit uses the Reflection API for obtaining the list of methods to run. This change in Reflection API caused tests run by JUnit to fail due to test-order dependencies.  See the test failure detected by the nondex tool by shuffling the java.lang.Class#getDeclaredFields() method.
image

The proposed fix preserves the order as it is even if the API changes its implementation. Hence, it removes the flakiness of the test.

vojtechhabarta:
It's true that `getDeclaredFields` method could theoretically return fields in any order. It is stated in specification (Javadoc)

> The elements in the returned array are not sorted and are not in any particular order.

But in fact it returns fields still in source order. Changing this behavior would have much bigger impact than few broken tests in this project.

For me having TypeScript properties in some logical order (same as in Java) is important feature. To achieve this typescript-generator project depends on current `getDeclaredFields` implementation. I think it is not even good to hide this by making tests resilient to this change. If this ever happens I want to know about it (failed tests).

Issue #472, opened 2020-04-02
Make all primitives non nullable?


mellson:
Is it possible to make all Java primitives non nullable without having to decorate an existing code base with annotations?

I have seen that requiredAnnotations have been added, but I don't think I can use that to say that I want all the primitives to be non null (without annotations). But maybe I'm wrong.

vojtechhabarta:
Currently this is not possible. If I understand it correctly you would like to have all properties optional except for those with primitive type and except for those annotated with some "required" annotations. Is that right?

mellson:
@vojtechhabarta that is correct!

vojtechhabarta:
I think this feature is implemented in #632 which was released in [2.30.840](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.30.840).

mellson:
@vojtechhabarta neat, thanks 😊

PR #632, opened 2021-02-11
Add parameters make all primitives non nullable

kk-zu:
Add `allPrimitivesRequired` parameter for make all primitives non nullable.(#472 )

I think that we use `requiredAnnotations` to make `String` or some `Class` treat as non nullable. But then primitive properties is created with nullable even though it's not nullable in Java.

So we use some annotation that treat as non nullable for primitive properties now even though it's not necessary in product.
 I want to solve that problem.

It is not breaking change. What do you think?

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -123,7 +123,7 @@ public class Settings {
     public List<Class<? extends Annotation>> optionalAnnotations = new ArrayList<>();
     public List<Class<? extends Annotation>> requiredAnnotations = new ArrayList<>();
     public List<Class<? extends Annotation>> nullableAnnotations = new ArrayList<>();
-    public boolean allPrimitivesRequired = false;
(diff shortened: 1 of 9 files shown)


kk-zu:
I'm sorry. There is a bug, so fix it.

kk-zu:
Sort java imports alphabetically. It is ok.

vojtechhabarta:
Thanks for your PR. I would like to request few changes:
- rename `allPrimitivesRequired` parameter to `primitivePropertiesRequired`
- instead of detecting primitive types by name add following method to `Utils` class and use it
``` java
    public static boolean isPrimitiveType(Type type) {
        if (type instanceof Class<?>) {
            final Class<?> cls = (Class<?>) type;
            return cls.isPrimitive();
        }
        return false;
    }
```

kk-zu:
Thanks for reviewing my PR. I fixed. Could you check some updated.

vojtechhabarta:
Released in [2.30.840](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.30.840).

PR #636, opened 2021-02-18
support mapping byte[] to string, fixes #462

### LINKED ISSUE #462: Byte array being mapped as any.
It's stated that byte should map to number, however the following java class the byte array gets mapped to any.  Is there a jackson2 annotation I'm missing?

``` java
public class BrowserFileItem implements BrowserFileItemInterface {
    private final Long id;
    private final byte[] data;
    private final String fileName;
    private final String fileExtension;
    private final String mimeType;

    public BrowserFileItem(final Long id, final byte[] data, final String fileName, final String fileExtension, final String mimeType) {
        this.id = id;
        this.data = data;
        this.fileName = fileName;
        this.fileExtension = fileExtension;
        this.mimeType = mimeType;
    }

    public Long getId() {
        return this.id;
    }

    public byte[] getData() {
        return this.data != null ? this.data.clone() : null;
    }

    public String getFileName() {
        return this.fileName;
    }

    public String getFileExtension() {
        return this.fileExtension;
    }

    public String getMimeType() {
        return this.mimeType;
    }
}
```

Becomes

``` typescript
    interface BrowserFileItemInterface {
        id: number;
        fileName: string;
        data: any;
        mimeType: string;
        fileExtension: string;
    }
```

panchenko:


vojtechhabarta:
Thank you. I will make the support for custom mapping of arrays more general.

vojtechhabarta:
580e37b60c7496e4b1d11eb0cf3773a54f324c8f

vojtechhabarta:
Released in [2.30.840](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.30.840).

PR #677, opened 2021-05-29
Fix some flaky tests in test class `EnumTest`

ailen-wrx:
Some of the tests in class EnumTest can fail due to the disorder of Enum objects, for example:
```python
[ERROR]   EnumTest.testSingleEnum:44 expected:<...ection = "North" | "[East" | "South" | "We]st";
> but was:<...ection = "North" | "[West" | "South" | "Ea]st";
>
```
Here's a list of all tests which fail due to this type of error:
```python
cz.habarta.typescript.generator.EnumTest.testEnumAsEnum
cz.habarta.typescript.generator.EnumTest.testEnumAsInlineUnion
cz.habarta.typescript.generator.EnumTest.testEnumAsNumberBasedEnum
cz.habarta.typescript.generator.EnumTest.testEnumAsUnion
cz.habarta.typescript.generator.EnumTest.testEnumMapKeys_asInlineUnion
cz.habarta.typescript.generator.EnumTest.testEnumUsingToString
cz.habarta.typescript.generator.EnumTest.testEnumWithJsonPropertyAnnotations
cz.habarta.typescript.generator.EnumTest.testEnumWithJsonValueFieldAnnotation
cz.habarta.typescript.generator.EnumTest.testEnumWithJsonValueMethodAnnotation
cz.habarta.typescript.generator.EnumTest.testSingleEnum
```
The reason is that according to the [Documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html), the method `Java.lang.Class.getDeclaredField()` returns an array containing Field objects reflecting all the accessible public fields of the class or interface represented by this Class object, while the elements in the array returned are **not sorted** and are **not in any particular order**. 

One way to fix this problem is to use the method `Java.lang.Class.getEnumConstants()`, which returns the elements in the declared order of this enum class.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
@@ -724,7 +724,6 @@ public class Jackson2Parser extends ModelParser {
                 try {
                     return enumClass.getDeclaredField(e.name());
                 } catch (NoSuchFieldException noSuchFieldException) {
-                    noSuchFieldException.printStackTrace();
                     return null;
                 }
             }).collect(Collectors.toList());
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
@@ -719,9 +719,15 @@ public class Jackson2Parser extends ModelParser {
 
         final List<EnumMemberModel> enumMembers = new ArrayList<>();
         if (sourceClass.type.isEnum()) {
-            final Class<?> enumClass = (Class<?>) sourceClass.type;
-            final Field[] allEnumFields = enumClass.getDeclaredFields();
-            final List<Field> constants = Arrays.stream(allEnumFields).filter(Field::isEnumConstant).collect(Collectors.toList());
+            final Class<Enum> enumClass = (Class<Enum>) sourceClass.type;
+            final List<Field> constants = Arrays.stream(enumClass.getEnumConstants()).map(e -> {
+                try {
+                    return enumClass.getDeclaredField(e.name());
+                } catch (NoSuchFieldException noSuchFieldException) {
+                    noSuchFieldException.printStackTrace();
+                    return null;
+                }
+            }).collect(Collectors.toList());
             for (Field constant : constants) {
                 Object value;
                 try {


typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
```diff
@@ -719,9 +719,14 @@ private DeclarationModel parseEnumOrObjectEnum(SourceType<Class<?>> sourceClass,
 
         final List<EnumMemberModel> enumMembers = new ArrayList<>();
         if (sourceClass.type.isEnum()) {
-            final Class<?> enumClass = (Class<?>) sourceClass.type;
-            final Field[] allEnumFields = enumClass.getDeclaredFields();
-            final List<Field> constants = Arrays.stream(allEnumFields).filter(Field::isEnumConstant).collect(Collectors.toList());
+            final Class<Enum> enumClass = (Class<Enum>) sourceClass.type;
+            final List<Field> constants = Arrays.stream(enumClass.getEnumConstants()).map(e -> {
```
panchenko, line 723: TBH, we don't actually need fields, the result of `enumClass.getEnumConstants()` can be used directly

ailen-wrx: As I saw in the following code, it seems that the `Field` object is still required for the methods `Field.getAnnotation` and `public EnumMemberModel(String propertyName, String enumValue, Field originalField, List<String> comments)`. It might be quite complex to use `Enum[]` directly...

vojtechhabarta: Yes, `Field` is actually needed for getting `@Deprecated` annotation.

vojtechhabarta:
@ailen-wrx thanks for your pull request.
I am interested where have you encountered this problem, I tried few different Java versions and haven't seen this problem.

vojtechhabarta:
Merged. I just changed exception handling (06b881e5f49b4d8487cdc299645023265778a402). NoSuchFieldException should never happed but if it does I would rather fail immediately than later (with NullPointerException).

vojtechhabarta:
Released in [v2.32.889](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.32.889).

ailen-wrx:
> @ailen-wrx thanks for your pull request.
> I am interested where have you encountered this problem, I tried few different Java versions and haven't seen this problem.

Thank you for accepting my PR. I have not encountered this problem on current JVM versions which return the fields in the order expected by the tests. I used the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, specifically `mvn -pl typescript-generator-core edu.illinois:nondex-maven-plugin:1.1.2:nondex`, and found that the tests would fail if the method `getDeclaredFields` returns the fields in the wrong order. Several more tests fail for similar reasons, and my teammates and I are preparing PRs for those. I hope it's fine with you if we open those PRs soon.

huseyingokay:
Hello,
I encountered the same problem by using the NonDex tool, specifically ```mvn -pl typescript-generator-core edu.illinois:nondex-maven-plugin:1.1.2:nondex``` in ImmutablesTest, and found that the test would fail if the method getDeclaredFields returns the fields in the wrong order. You can check out my PR: https://github.com/vojtechhabarta/typescript-generator/pull/686

vojtechhabarta:
The problem with `getDeclaredFields` is that in real life it always returns fields in source-code order and people count on this behavior. I think usually it is just for providing properties in some logical order (for better readability) and doesn't affect functionality but even in this case the order is important. In Jackson it is possible to specify order using `@JsonProperty.index` or `@JsonPropertyOrder` but that means extra work and it is error-prone.

So I don't want to change **tests** which depend on `getDeclaredFields` order because I would really like to see those test failing if some future Java version changes the behavior this method. However I would fix such problems in **production** code. And I would fix other problems (like order of entries in `HashMap`) in both production and test code.

ailen-wrx:
> The problem with `getDeclaredFields` is that in real life it always returns fields in source-code order and people count on this behavior. I think usually it is just for providing properties in some logical order (for better readability) and doesn't affect functionality but even in this case the order is important. In Jackson it is possible to specify order using `@JsonProperty.index` or `@JsonPropertyOrder` but that means extra work and it is error-prone.
> 
> So I don't want to change **tests** which depend on `getDeclaredFields` order because I would really like to see those test failing if some future Java version changes the behavior this method. However I would fix such problems in **production** code. And I would fix other problems (like order of entries in `HashMap`) in both production and test code.

Thank you for your follow-up! You're absolutely right that `getDeclaredFields` has so far returned fields in the order in the `.class` file (which itself is the order in the `.java` file when compiling with `javac`). However, `getDeclaredFields` does not guarantee to remain like that, e.g., `getDeclaredMethods` used to be like that until Java 6 and then changed in Java 7, breaking a lot of JUnit tests that relied on the implicit test order. (You also mention `get(Declared)Methods` in [#686](https://github.com/vojtechhabarta/typescript-generator/pull/686).) Also, [the documentation of `enumMembers`](https://docs.oracle.com/javase/tutorial/reflect/special/enumMembers.html) emphasizes: 

> *Note*: For various reasons, including support for the evolution of the enum type, the declaration order of enum constants is important. `Class.getFields()` and `Class.getDeclaredFields()` do not make any guarantee that the order of the returned values matches the order in the declaring source code. If ordering is required by an application, use `Class.getEnumConstants()`.

We'll follow your guide to see if we can reasonably change production code (in `src`) in `typescript-generator-core` (or other modules) to address some flaky tests. If we find some idea, we'll open a PR so you can tell us if it looks good to you. If you have some other suggestions where you think that some (motivated but inexperienced! :)) undergraduate students can help your project, please let us know. Thanks again.

PR #686, opened 2021-06-02
Flaky test is fixed by adding JsonPropertyOrder

huseyingokay:
- The `testImmutables` method in `ImmutablesTest` can fail if the order of fields in `getDeclaredField` changes. I found this issue using the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool.

* The command ```mvn -pl typescript-generator-core edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=cz.habarta.typescript.generator.ImmutablesTest#testImmutables -DnondexRuns=10``` runs the test 10 times and results in 5 failures in my JVM.

- After the executions, there was only a problem that caused all the errors: the width and height properties were replacing in the result string. For instance, the result can contain:

     "interface Rectangle extends Shape {\n" +
                    "    kind: 'rectangle';\n" +
                    "    width: number;\n" +
                    "    height: number;\n" +
                "}\n"   or,

     "interface Rectangle extends Shape {\n" +
                    "    kind: 'rectangle';\n" +
                    "    height: number;\n"  +
                    "    width: number;\n" +
                "}\n"

- To prevent such failures, the order of the properties could be specified using JsonPropertyOrder. 

-  After this change, the test passed in all 10 runs with ```mvn -pl typescript-generator-core edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=cz.habarta.typescript.generator.ImmutablesTest#testImmutables -DnondexRuns=10```.

### DIFF
--- a/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/ImmutablesTest.java
+++ b/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/ImmutablesTest.java
@@ -1,6 +1,7 @@
 
 package cz.habarta.typescript.generator;
 
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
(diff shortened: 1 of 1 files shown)


vojtechhabarta:
Thanks for very good description. I also tried it myself.

Please see https://github.com/vojtechhabarta/typescript-generator/pull/677#issuecomment-853363371. However in this case I am not sure if the problem is caused by `getDeclaredFields` method (which returns fields in source-code order in the real life) or by `getDeclaredMethods` or `getMethods` (which are really "random") so I will merge your change.

PR #687, opened 2021-06-11
Run node tests from maven

olegshtch:


vojtechhabarta:
Thanks for this PR.
I will merge it but it slows down first build significantly so I will make it run only when some profile is activated.

PR #706, opened 2021-08-22
Add ability to use record types instead of index signatures

crummy:
My team uses `Record<string, number>` syntax instead of `{ [index: string]: number }` syntax. Here is a PR to allow use of this syntax instead. Not very experienced in this codebase so I'm open to feedback and changes.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java
@@ -140,6 +140,7 @@ public class Settings {
     public boolean jackson2ModuleDiscovery = false;
     public List<Class<? extends Module>> jackson2Modules = new ArrayList<>();
     public ClassLoader classLoader = null;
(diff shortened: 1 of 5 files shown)


vojtechhabarta:
Your approach would work but technically `Record<string, number>` is not indexed type formatted differently, it is generic type for which there is already `GenericBasicType`.

So more correct solution is to transform `IndexedArrayType` to `GenericBasicType` in `ModelCompiler`. You can create method in `ModelCompiler` where you can reuse `transformBeanPropertyTypes` method and call it after `transformOptionalProperties`.

Another solution is to create an `Extension` and do the transformation there. The extension could be merged to typescript-generator if you think it would be interesting for others as well.

vojtechhabarta:
Feature released in [v2.36.1070](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.36.1070) (different implementation).
In Maven use `<mapMap>asRecord</mapMap>` parameter.

PR #709, opened 2021-08-31
Export function uriEncoding

ArthurFiorette:
When it is necessary to request an object, use a cache for example, for a certain path, if the encoded path is different, the result will be different, so it is necessary to handwrite this exactly function elsewhere and export it.

It shouldn't matter to anyone who isn't using it, but to whoever, it's nice to have.


vojtechhabarta:
Proper support for exporting this function would require more lines of code than the function has.

PR #716, opened 2021-09-10
Bump JUnit version to 5.7.2

olegshtch:


vojtechhabarta:
I realized that after JUnit upgrade to version 5 tests in `typescript-generator-spring` module were not running with message _"TestEngine with ID 'junit-jupiter' failed to execute tests"_ and with `ClassNotFoundException`. I am not sure what was the problem but changing dependencies (d430f80ae25c734b58f9b137f3d2c79693e87334) solved it.

PR #688, opened 2021-06-24
Fix a number of flaky tests

mohasarc:
Based on [your comment](https://github.com/vojtechhabarta/typescript-generator/pull/677#issuecomment-853363371) and the follow-up comment to change `src/main`, I'm opening this PR to address some tests by changing `src/main`. If you prefer to address this differently, please let me know so I can revise the PR.

According to the test results obtained using the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool, I have confirmed several flaky tests; these tests can pass or fail non-deterministically. The main reason for this flakiness is the order of the fields returned by the `getDeclatedFields()` method. This method has been used internally by Jackson library's  `ObjectMapper`; therefore, the default output of `ObjectMapper` is not deterministic [(documentation)](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/MapperFeature.html#SORT_PROPERTIES_ALPHABETICALLY). 

https://github.com/mohasarc/typescript-generator/blob/107eb2f854692bd1ad466596f62b21fcc3fc735c/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java#L116

To solve this, I have configured `ObjectMapper` to return the fields in alphabetical order. 

Finally, I have changed some of the tests into an alphabetical order to fit the changes to the main code. here is a list of all the flaky tests that were fixed:

- `cz.habarta.typescript.generator.EnumTest.testObjectEnum`
- `cz.habarta.typescript.generator.GenericsTest.testAdvancedGenerics`
- `cz.habarta.typescript.generator.JsonUnwrappedTest.test`
- `cz.habarta.typescript.generator.JsonUnwrappedTest.testPrivateField`
- `cz.habarta.typescript.generator.ClassesTest.testUnsortedConstructor`
- `cz.habarta.typescript.generator.Jackson2ParserTest.test`
- `cz.habarta.typescript.generator.OptionalTest.test`
- `cz.habarta.typescript.generator.OptionalTest.testDeclarationNullableAndUndefinableType`
- `cz.habarta.typescript.generator.OptionalTest.testDeclarationNullableType`
- `cz.habarta.typescript.generator.OptionalTest.testDeclarationQuestionMark`
- `cz.habarta.typescript.generator.OptionalTest.testDeclarationQuestionMarkAndNullableType`
- `cz.habarta.typescript.generator.OptionalTest.testDeclarationUndefinableType`
[...]

PR #717, opened 2021-09-22
Fixed flakiness in MapEntryTest#testDefaultShapes

Augustine-C:
This test is flaky since the method `java.lang.Class.getDeclaredFields` is non-deterministic, and `GetDeclaredFilelds`  is used in several methods in the `com.fasterxml.jackson` library (e.g. `objectMapper.writeValueAsString`). 
I changed the test to compare two jsonNode objects generated from the strings, so the specific order of the properties would not affect the result. 
Also, I added checks for both variants of the generated typescript string due to the non-deterministic ordering of properties in the output. 
The other two tests in `MapEntryTest.java` contain the same issues and can be fixed with similar changes in this PR.


vojtechhabarta:

PR #718, opened 2021-09-25
fix flaky test for Jackson1Parser

CharlesZKQ:
I found a flaky test (test can fail occasionally) in `cz.habarta.typescript.generator.Jackson1ParserTest`. The reason why this test is flaky is that when` Jackson1Parser`  is initialized, it will use  `ObjectMapper ` to parse and construct the model, the `ObjectMapper ` is random since it used `getDeclatedFields()` and `ConcurrentHashMap` which is not deterministic. I fixed it by changed the attribute of `SORT_PROPERTIES_ALPHABETICALLY` to true, that guarantees the order of all properties parsed by `Jackson1Parser`


vojtechhabarta:

PR #719, opened 2021-09-25
fix flaky test for testJackson

CharlesZKQ:
I found a flaky test (test can fail occasionally) in `cz.habarta.typescript.generator.ObjectAsIdTest.testJackson`. The reason for its flakiness is the same as the last PR I submitted. the `ObjectMapper `is not deterministic. I fixed it by configure the `ObjectMapper` in `getObjectMapper() `.


vojtechhabarta:

PR #720, opened 2021-09-26
fix flaky test in GsonParserTest

CharlesZKQ:
I found a flaky test in `cz.habarta.typescript.generator.parser.GsonParserTest#test`. The reason why this test is flaky is that `GsonParser `is non-deterministic, it used `getDeclatedFields()` for constructing `Gson `oject. Since it doesn't have direct configuration to set properties' order, I chose to sort the property list by their name, which is not optimal though, it fixed the flakiness of this test.


vojtechhabarta:

PR #721, opened 2021-09-28
Check exclude filter when parsing spring controller methods

jpstratman:
We're running into a scenario where some mock controller methods are being included that we'd rather exclude. Currently there doesn't seem to be a way to exclude individual controller methods (or even individual controllers?) via the excludeClassPattern setting, so this is an attempt to add that functionality.

Just let me know if this functionality already exists in some form or if this is the wrong approach and I can make any needed changes 😄

### DIFF
--- a/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
+++ b/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
@@ -186,7 +186,11 @@ public class SpringApplicationParser extends RestApplicationParser {
         final List<Method> methods = getAllRequestMethods(controllerClass);
         methods.sort(Utils.methodComparator());
         for (Method method : methods) {
-            parseControllerMethod(result, context, controllerClass, method);
+            // Do not use default Method.toString since that will include parameter types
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Excluding individual controllers is possible using [excludeClassPatterns](https://www.habarta.cz/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html#excludeClassPatterns). (If your controller is nested class you just need to use `$` character.)

Unfortunately excluding individual methods in Spring controller is not possible. In JAX-RS resource it is possible to use Swagger annotation `@ApiOperation` with `hidden` set to `true`.
https://github.com/vojtechhabarta/typescript-generator/blob/64caab5012fb58c1bef3ab0636d156a29830c21f/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/JaxrsApplicationParser.java#L161-L163
Would it be interesting for you to add support for such annotation also for Spring controllers?

Or maybe we could introduce parameter `excludeMethodAnnotations` that would allow to use custom annotations for excluding controller methods (similarly to `excludePropertyAnnotations`) parameter for properties.

I am suggesting different ways because reusing class exclusion parameters for methods doesn't seem good to me.

vojtechhabarta:
Support for Swagger annotations release in [v2.33.956](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v2.33.956).

PR #730, opened 2021-10-15
Add test to check data generated from Java in TypeScript and vice versa

olegshtch:


vojtechhabarta:
It should be possible to reset `excludedGroups` property back to "empty" value using `combine.self="override"` like this:
``` xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
        <excludedGroups combine.self="override"></excludedGroups>
    </configuration>
</plugin>
```
Do you want to try it?

olegshtch:
@vojtechhabarta I'm not sure how to call it twice and second run after npm tests.

vojtechhabarta:
I think it should be possible to run it by adding second `maven-surefire-plugin` execution after `npm test`:
``` xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <executions>
        <execution>
            <id>test-typescript2java</id>
            <phase>test</phase>
            <goals>
                <goal>test</goal>
            </goals>
            <configuration>
                <excludedGroups combine.self="override"></excludedGroups>
                <groups>typescript2java</groups>
            </configuration>
        </execution>
    </executions>
</plugin>
```

panchenko:
TBH, I am not sure why mixing the things. The generated JSON can be tested in the same language.

olegshtch:
@vojtechhabarta Looks like it works.

@panchenko It's a project for mixing TypeScript and Java so I suppose it better to have same tests.

vojtechhabarta:
@panchenko I don't want to complicate build but this is activated using profile and might be useful.

PR #731, opened 2021-10-15
Check exclude filter when parsing spring controller classes by annotation

jpstratman:
Adding class exclusion functionality in `SpringApplicationParser` when `scanSpringApplication` is not true (class exclusion is already working when `scanSpringApplication` is true)

### DIFF
--- a/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
+++ b/typescript-generator-spring/src/main/java/cz/habarta/typescript/generator/spring/SpringApplicationParser.java
@@ -121,7 +121,7 @@ public class SpringApplicationParser extends RestApplicationParser {
         // controller
         final Component component = AnnotationUtils.findAnnotation(cls, Component.class);
         if (component != null) {
-            if (isClassNameExcluded != null && isClassNameExcluded.test(sourceType.toString())) {
+            if (isClassNameExcluded != null && isClassNameExcluded.test(cls.getName())) {
                 return null;
             }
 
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
BTW build is failing because in `main` branch JUnit 4 was upgraded to version 5 where `Assert` was replaced by `Assertions`.

jpstratman:
> The class exclusion is done before the list of classes is passed to `generateTypeScript` method. When I try `sample-maven-spring` project and configure `<excludeClassPatterns>` it works:
> 
> ```
> <configuration>
>     ...
>     <classPatterns>
>         <pattern>cz.habarta.typescript.generator.sample.spring.**</pattern>
>     </classPatterns>
>     <excludeClassPatterns>
>         <pattern>cz.habarta.typescript.generator.sample.spring.SpringTestApplication</pattern>
>     </excludeClassPatterns>
>     <scanSpringApplication>false</scanSpringApplication>
>     <generateSpringApplicationClient>true</generateSpringApplicationClient>
> </configuration>
> ```
> 
> Do you think it is needed to apply exclusion filter also in `SpringApplicationParser`?

I tried that out as well, but the generated spring endpoint class still contained the SpringTestApplication methods even though it correctly excluded the type interfaces from that class. From what I can tell, there is a check needed in SpringApplicationParser to get that to work. After the changes in this PR, it is excluding that class from the generation correctly.

vojtechhabarta:
I am still not sure. SpringTestApplication is not a controller and doesn't have any REST method. Do you mean `SpringTestApplication.GreetingController`? If I try mentioned configuration then generated `RestApplicationClient` contains just `greeting` method which I think is correct. I want to understand your point better because it might be also relevant to JAX-RS applications.

PR #749, opened 2021-11-18
Create .travis.yml

suzzy777:


vojtechhabarta:
Typescript-generator is built using AppVeyor and I don't want to maintain two build systems so I will close this PR.
If you think Travis should be used for some reason please comment.

PR #762, opened 2021-12-05
Support for Jakarta JSONB version 2

vojtechhabarta:
This PR adds support for Jakarta JSON bind version 2. To have support for both version 1 and version 2 it was needed to use `javax.json.bind:javax.json.bind-api` dependency for version 1 and `jakarta.json.bind:jakarta.json.bind-api` for version 2.

I didn't want to copy&paste implementation so I changed `JsonbParser` to support both versions at the same time. Also `JsonbParserTest` is reused. It was updated for Jakarta and is copied in build time with replaced `jakarta.json.bind` imports to `javax.json.bind`. (It is done in a similar way as support for JAX-RS version 2 (javax) and version 3 (jakarta) in `JaxrsApplicationParser`.)


vojtechhabarta:
@rmannibucau you are the author of original implementation so if you want to comment I will leave this PR open for a week.

rmannibucau:
Works for me. I'd just mention i tend to use maven shade plugin relocation to do it and a dedicated module but it is an impl detail, no user impact so let's go.

vojtechhabarta:
Thanks you. Merging.

PR #775, opened 2022-01-06
feature: add the ability to remove the timestamp from the file comment

hagai-reuveni:
## What is the problem?
At the moment, if one is using the generator in source control and using a build tool, there is a problem with the generated timestamp at the file comment (different in every run).
To avoid the above problem one can remove the entire comment (`noFileComment=true`) but that means losing other data on that statement ("_generated using typescript-generator version_") which might be valuable.

note: the `noFileComment` is false by default

## The fix
Add a `noTimestampInFileComment` prop that when set to true will omit the timestamp from the top of the file comment.


vojtechhabarta:
This would be second parameter for the same purpose.
I think it should be possible to use `noFileComment` parameter and then add any text by build tool if really needed.

hagai-reuveni:
@vojtechhabarta 
The problem is not with the file comment but with having a timestamp embedded in the file.
What do you think about just removing the timestamp from the comment?

PR #798, opened 2022-02-13
support multiple names for jackson subtype

### LINKED ISSUE #796: @JsonSubTypes.Type names are neglected, only the first name is emitted
Jackson supports specifying `names` for subtypes, so the following is possible
```java
@JsonTypeInfo(
        use = JsonTypeInfo.Id.NAME,
        include = JsonTypeInfo.As.EXISTING_PROPERTY,
        property = "type",
        visible = true
)
@JsonSubTypes({
        @JsonSubTypes.Type(value = LongDto.class, names = {"Price", "Duration"}),
        @JsonSubTypes.Type(value = BoolDto.class, names = {"Official", "Signer"}),
})
static class AbstractDto {
    public String type;
}

static class LongDto extends AbstractDto {
    public long value;

    @Override
    public String toString() {
        return "LongDto{" +
                "type='" + type + '\'' +
                ", value=" + value +
                '}';
    }
}

static class BoolDto extends AbstractDto {
    public boolean value;

    @Override public String toString() {
        return "BoolDto{" +
                "type='" + type + '\'' +
                ", value=" + value +
                '}';
    }
}

@Test
void test() throws JsonProcessingException {
    LongDto priceDto = new LongDto();
    priceDto.type = "Price";
    priceDto.value = 100;
    LongDto durationDto = new LongDto();
    durationDto.type = "Duration";
    durationDto.value = 2;

    BoolDto boolDto = new BoolDto();
    boolDto.type = "Official";
    boolDto.value = true;

    ObjectMapper objectMapper = new ObjectMapper();
    String jsonString = objectMapper.writeValueAsString(List.of(priceDto, boolDto, durationDto));
    System.out.println(jsonString); // [{"type":"Price","value":100},{"type":"Official","value":true},{"type":"Duration","value":2}]
    System.out.println(objectMapper.readValue(jsonString, new TypeReference<List<AbstractDto>>() { }));  // [LongDto{type='Price', value=100}, BoolDto{type='Official', value=true}, LongDto{type='Duration', value=2}]
}

```

I would expect the generator to emit something like this:
```ts
export interface AbstractDto {
  'type': 'Price' | 'Duration' | 'Official' | 'Signer';
}

export interface BoolDto extends AbstractDto {
  type: 'Official' | 'Signer';
  value: boolean;
}

export interface LongDto extends AbstractDto {
  type: 'Price' | 'Duration'';
  value: number;
}

export type AbstractDtoUnion = BoolDto | LongDto;
```

But the actual output is like this (only first name is taken of each type):
```ts
export interface AbstractDto {
  type: 'Official' | 'Price';
}

export interface BoolDto extends AbstractDto {
  type: 'Official';
  value: boolean;
}

export interface LongDto extends AbstractDto {
  type: 'Price';
  value: number;
}

export type AbstractDtoUnion = LongDto | BoolDto;
```

is it possible to support this case?

andrewbents:
fixes #796 

resolves all subtype names instead of only the first one

### DIFF
--- a/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/Jackson2ParserTest.java
+++ b/typescript-generator-core/src/test/java/cz/habarta/typescript/generator/Jackson2ParserTest.java
@@ -97,10 +97,10 @@ public class Jackson2ParserTest {
         Assertions.assertNull(bean2.getTaggedUnionClasses());
         Assertions.assertNull(bean3.getTaggedUnionClasses());
         Assertions.assertEquals("kind", bean0.getDiscriminantProperty());
-        Assertions.assertEquals("explicit-name1", bean1.getDiscriminantLiteral());
-        Assertions.assertEquals("SubType2", bean2.getDiscriminantLiteral());
-        Assertions.assertEquals("Jackson2ParserTest$SubTypeDiscriminatedByName3", bean3.getDiscriminantLiteral());
-        Assertions.assertEquals("Jackson2ParserTest$SubTypeDiscriminatedByName4", bean4.getDiscriminantLiteral());
+        Assertions.assertEquals(Arrays.asList("explicit-name1", "SubType1"), bean1.getDiscriminantLiterals());
+        Assertions.assertEquals(Arrays.asList("SubType2"), bean2.getDiscriminantLiterals());
+        Assertions.assertEquals(Arrays.asList("Jackson2ParserTest$SubTypeDiscriminatedByName3"), bean3.getDiscriminantLiterals());
+        Assertions.assertEquals(Arrays.asList("Jackson2ParserTest$SubTypeDiscriminatedByName4"), bean4.getDiscriminantLiterals());
     }
 
     @Test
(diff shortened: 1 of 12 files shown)


typescript-generator-core/src/test/java/cz/habarta/typescript/generator/Jackson2ParserTest.java
```diff
@@ -132,7 +132,7 @@ static Jackson2Parser getJackson2Parser() {
 
     @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
     @JsonSubTypes({
-        @JsonSubTypes.Type(value = SubTypeDiscriminatedByName1.class, name = "SubType1"), // value from @JsonTypeName is used
+        @JsonSubTypes.Type(value = SubTypeDiscriminatedByName1.class, name = "SubType1"),
```
andrewbents, line 135: not sure about this one

this is probably a breaking change, since @JsonTypeName would no longer have priority over @JsonSubTypes.Type, and all typenames would be included

PR #801, opened 2022-02-20
Recognize Swagger Api annotation as RestController contract

thilken:
-adapt implementation to recognize Swagger Api annotation as RestController contract so the generator is compatible with current openapi-generator (https://github.com/openapitools/openapi-generator)


vojtechhabarta:
Does Spring also recognize `io.swagger.annotations.Api` annotation (in addition to `org.springframework.stereotype.Component`) or is it meant just for Swagger?

PR #829, opened 2022-04-20
Add `usePeerDeps` option to module dependency.

kschaefe:
Normal dependencies can cause type mismatches if the dependency versions
are different.  This means that our dependency and every other library
or application depending on the same dependency need to use compatible
versions.  This is where `peerDependencies` is better.  The dependency
is only managed once at the top-level.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ModuleDependency.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/ModuleDependency.java
@@ -15,25 +15,27 @@ public class ModuleDependency {
     public File infoJson;
     public String npmPackageName;
     public String npmVersionRange;
+    public boolean usePeerDeps;
 
     public ModuleDependency() {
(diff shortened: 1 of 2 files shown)


kschaefe:
I don't really know much about Maven plugins, so I only touched the absolute smallest amount of code that I could.  I think the result is right, if the approach is not.  `peerDependencies` is a better way to manage the dependencies of these linked generated libraries.

kschaefe:
Any feedback on this or chance to get it included?  I now have a build failing because this plugin is using `dependencies` instead of `peerDependencies`.  `tsc` is failing to do proper module resolutions because of this.

vojtechhabarta:
Thanks for your PR.
I just renamed `usePeerDeps` to `peerDependency`.

vojtechhabarta:
Released in [3.2.1263](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v3.2.1263).

PR #846, opened 2022-05-25
add retention policy check to optional annotations

serv-inc:
Thanks for the typescript-generator. This is a small PR to check if all optional annotations are usable. It might be extended to the other annotation types.

Using e.g. org.jetbrains.annotation.nullable fails silently, as it has RetentionPolicy.class. This check fails if an annotation with this retention policy is used for optional annotations.


vojtechhabarta:
I've just added similar check for all annotations. It is now in `validate()` method as you suggested. Thanks.

Issue #856, opened 2022-07-28
Typescript generator cannot handle Java-Records


digital-h:
Hi,

as far as I can see, the typescript-generator unfortunately cannot handle Java-records adequately.

If I have a standard Java-Class such as the following:

`
public class Foo {

	private final String bar;

	public Foo(String bar) {
		this.bar = bar;
	}

	public String getBar() {
		return bar;
	}
}
`

Then a correct Typescript is generated:

`
/* tslint:disable */
/* eslint-disable */
// Generated using typescript-generator version 2.36.1070 on 2022-07-28 16:58:23.

export class Foo {
    bar: string;

    constructor(data: Foo) {
        this.bar = data.bar;
    }
}
`

But as soon as I change this to a java record like so

`
public record Foo(String bar) {
}
`

Then the constructor receives a super(data), which results into a typescript error, because there is no superclass

`
/* tslint:disable */
/* eslint-disable */
// Generated using typescript-generator version 2.36.1070 on 2022-07-28 16:59:06.

export class Foo {
    bar: string;

    constructor(data: Foo) {
        super(data);   <---- THIS CAUSES AN ERROR!!!!
        this.bar = data.bar;
    }
}
`

My generator is configured as follows:

`
<jsonLibrary>jackson2</jsonLibrary>
<outputKind>module</outputKind>
<outputFileType>implementationFile</outputFileType>
<mapClasses>asClasses</mapClasses>
<generateConstructors>true</generateConstructors>
`

Is this a bug or am I missing some configuration?

ThomasVitale:
It would be really great to support Java 17+ and records in particular

vojtechhabarta:
@digital-h thanks for detailed issue. There was a bug which affected constructors in cases where parent class was excluded (which is also the case with records).

@ThomasVitale supported Java versions are now 11, 17 (including records) and newer. There was just mentioned bug (not specific to records).

staeff777:
Is it possible, that you can trigger a release with this bugfix? It would be useful in a current project.

vojtechhabarta:
Released in [3.1.1185](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v3.1.1185).

PR #857, opened 2022-07-29
Do not treat java.lang.Record as superclass to prevent a super()-call…

digital-h:
… in the constructor.

See: https://github.com/vojtechhabarta/typescript-generator/issues/856

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,7 +54,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>1.8</java.version>
+        <java.version>16</java.version>
         <kotlin.version>1.6.20</kotlin.version>
         <github.global.server>github</github.global.server>
     </properties>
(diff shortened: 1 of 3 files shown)


digital-h:
Unfortunately the pipeline still runs on java 8, so it cannot handle a higher language level. Any plans to support this in the future?

vojtechhabarta:
Minimal supported Java version is now 11 (there are many projects still using Java 11) so the build also runs on Java 11.
But records were already supported even when build was running on Java 8. There was just the bug with constructors you encountered.

I think I fixed it by this commit: 6c50e2164c8c2b3bfaf18c73372ee68e7cc78f7a - `parentType` in a `TsBeanModel` should be `null` when parent class is excluded but it was `Any` - [ModelCompiler.java](https://github.com/vojtechhabarta/typescript-generator/blob/main/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/compiler/ModelCompiler.java#L341). BTW `Record` super-class is excluded here: [Settings.java](https://github.com/vojtechhabarta/typescript-generator/blob/main/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/Settings.java#L665).

Thanks

PR #861, opened 2022-08-03
Feature: allow nullable constructor parameter

digital-h:
We are working together with the library https://github.com/typestack/class-transformer
This library does not provide all required values with the constructor, but instead sets the values after instantiation of the class.
Unfortunately this is not compatible, how typescript-generator works as default.
This PR adds the flag `allowNullableConstructorParameter` to the settings. By default this is false and keeps the current behaviour. If set to true a null-check is added the the constructor assignments.


vojtechhabarta:
I think you can achieve what you need differently:

Let's say we have some class in `clazz` variable: `let clazz: new () => any = ...`
then instead of using `new clazz()` which calls the constructor we can create instance using `Object.create(clazz.prototype)` which doesn't call the constructor.

PR #862, opened 2022-08-19
add support for JSON-B built-in types in the conversion

rmannibucau:
Concretely `JsonValue` children are not supported as of today leading to these types being treated as standard Java types (and they are not in regard of the serialization).
This PR just set some replacement for types to ensure the generator can treat them as expected.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,16 @@
                 <version>2.1.1</version>
             </dependency>
             <dependency>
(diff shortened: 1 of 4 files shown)


vojtechhabarta:
Released [3.2.1263](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v3.2.1263).

PR #883, opened 2022-09-06
fix mapPackagesToNamespaces in AxiosClientExtension

agaengel:
Wenn setting mapPackagesToNamespaces to true the generated typescript file has an error when using the AxiosClientExtension cause the client uses the class without the namespace


vojtechhabarta:
Released in [3.2.1263](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v3.2.1263).

PR #937, opened 2023-01-17
Support for sealed classes containing tagged unions

Nthalk:
This uses the proposal of https://github.com/FasterXML/jackson-module-kotlin/issues/239 and the general solution of https://github.com/FasterXML/jackson-module-kotlin/pull/240 to import subtypes of a sealed object or interface that are annotated with `@JsonTypeName` and assignable to the container class or interface to create a much less verbose method of generating tagged unions.

Sealed types are supported in kotlin 1.6+ and Java 17+, however, we don't really need the sealed functionality, as we are an offline, one-shot generator, this may work to generate the correct type mappings, but fail to parse on the runtime object mapper side unless the user is using the supported kotlin and java versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ pom.xml.releaseBackup
 pom.xml.versionsBackup
 pom.xml.next
 release.properties
+/.mvn/wrapper/maven-wrapper.jar
 
 # Gradle
 /sample-gradle/.gradle/
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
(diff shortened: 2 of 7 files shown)


vojtechhabarta:
Hello @Nthalk, I checked your PR and I like the idea of using sealed classes for tagged unions instead of `@JsonSubTypes`. However your implementation only works for classes declared inside tagged union root class while ObjectMapper could be able to handle subclasses declared anywhere. So in c716d4f32211b6459ca11ee6bde1f27f22171580 I implemented this feature by getting list of subclasses directly from Jackson which can be configured with appropriate module for handling sealed classes.

I understand your use case of one-shot generator but I see value in generating TypeScript code corresponding as much as possible to runtime behaviour of ObjectMapper (with added modules). For your use case you just need to add Jackson module handling sealed classes.

vojtechhabarta:
Here is an example of tagged union defined using Kotlin sealed class:

``` kotlin
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
// @JsonSubTypes(
//     JsonSubTypes.Type(Rectangle::class),
//     JsonSubTypes.Type(Circle::class),
// )
sealed class Shape {
}

@JsonTypeName("rectangle")
class Rectangle(val a: Int, val b: Int) : Shape() {
}

@JsonTypeName("circle")
class Circle(val radius: Int) : Shape() {
}
```

It is needed to add `com.fasterxml.jackson.module.kotlin.KotlinModule` to typescript-generator configuration parameter `jackson2Modules`.

In this case typescript-generator recognizes tagged union root class by `@JsonTypeInfo` annotation and there is no need for `@JsonSubTypes`.

For Java it is similar but currently there is no publicly available Jackson module for Java sealed classes. You can check example of such module in https://github.com/vojtechhabarta/typescript-generator/commit/c716d4f32211b6459ca11ee6bde1f27f22171580#diff-58156e5b6e0095f0b6f2173780b20ec0d7474cfca48257a44c68da1d75a0ef54, unfortunately it is commented out because typescript-generator build needs to run on Java 11.

Nthalk:
Awesome, I'll close this PR out then. When can we expect the next release? =D

vojtechhabarta:
Just released as part of [3.2.1263](https://github.com/vojtechhabarta/typescript-generator/releases/tag/v3.2.1263).

PR #946, opened 2023-02-05
Feature/inherited json identity info

kschaefe:
Jackson serializes subclasses of classes annotated with `JsonIdentityInfo` as if the class itself were annotated with `JsonIdentityInfo`.  See the updates to `ObjectAsIdTest.testJackson` to prove this is the case.  The TypeScript generator was not producing correct interface definitions because it was not properly recognizing all `JsonIdentityInfo` cases.  `JsonIdentityInfo` isn't `Inherited`, even if it were `Class.getAnnotation` doesn't return inherited annotations.  We need to manually walk the target class upward looking for `JsonIdentityInfo`.

This PR adds an example of this problem to the test cases.  First making the test cases pass with the current behavior.  Next making the test cases fail using the expected behavior.  Finally updating `Jackson2Parser` to make the tests pass.

I simply added a utility method in `Jackson2Parser` to support this change.  I see you have some utilities classes, but was unsure of your style for that.  The method should be easy to move wherever you feel is best.  Thanks.

### DIFF
--- a/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
+++ b/typescript-generator-core/src/main/java/cz/habarta/typescript/generator/parser/Jackson2Parser.java
@@ -367,6 +367,20 @@ public class Jackson2Parser extends ModelParser {
         }
     }
 
+    private <T extends Annotation> T findClassAnnotation(Class<?> cls, Class<T> annotation) {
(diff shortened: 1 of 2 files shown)


vojtechhabarta:
Thanks for your PR.
I think this annotation propagates not only from superclasses but also from interfaces. For that there is `getAnnotationRecursive` method, so I used it. Let me know if that's not correct.

PR #953, opened 2023-03-05
- Added settings.includePropertyFunctionImpl

dcrousset:
Hello sir,

Let me suggest another way to allow or not a property to be included just by using simple Function.  

If you need a more complete work, please let me know.

Thank you for your amazing generator :o)


dcrousset:
Sorry, this should be 2 differents pullrequests, but I don't know how to split these two new features.

PR #1006, opened 2023-09-04
Configuration Cache fix for GenerateTask

IgorDomagala:
Using typescript-plugin (v3.2.1263) with configuration cache enabled results in error:

```type `cz.habarta.typescript.generator.gradle.GenerateTask`: cannot serialize object of type 'org.gradle.api.tasks.compile.JavaCompile', a subtype of 'org.gradle.api.Task', as these are not supported with the configuration cache.``` 

and many more, mostly because GenerateTask uses variable "project" directly.

You can reproduce this error by running `./gradlew --configuration-cache --rerun-tasks assemble` .

This pull request fixes it by changing way of class loading and getting classpath.
This PR also adds Gradle configuration cache test.

### CODE REVIEW

**Comment by vojtechhabarta** [comment]
This project currently supports Gradle version 5.6 but this PR is using Gradle 8. I think it is not reasonable to break compatibility for people who are still using older versions of Gradle. I would consider updating minimal Gradle version to 6.

**Comment by IgorDomagala** [comment]
Hi I changed it to match your requirements ;)

**Comment by rouazana** [comment]
Hello, would you consider merging this PR?

**Comment by rouazana** [comment]
Hello @IgorDomagala 

Thank you very much for your contribution!

After some testing it seems your PR is not compatible with `customTypeMappings`.

For example with this configuration
```
        customTypeMappings = listOf("com.hopwork.model.currency.CurrencyCode:CurrencyCode")
        customTypeAliases = listOf("CurrencyCode:string")
```

I get the following error:

```
Running TypeScriptGenerator version 3.2-SNAPSHOT-main-1006
Loading class javax.annotation.Nullable
Loading class com.hopwork.model.currency.CurrencyCode

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:generateTypeScript'.
> Failed to parse configured custom type mapping 'com.hopwork.model.currency.CurrencyCode:CurrencyCode': java.lang.ClassNotFoundException: com.hopwork.model.currency.CurrencyCode
```

Do you have an idea of what could cause this?

PR #1059, opened 2024-04-22
Add support for controlled trailing semi-colons and injecting custom properties

manikandan-ravikumar:
In this PR, I would like to introduced the following two new settings,
1. `disableTrailingSemiColon` - A boolean flag indicating if the generator should append every line with a semi-colon
2. `injectCustomProperties` - A map of class to string property lists. Similar to `importDeclarations` but instead injects additional properties into the given type. This is useful for when we try to be backward-compatible in the FE API and update the backend logic to have the newer property names. Also avoids adding these properties manually every time after running the `typscript-generator`.

This PR also includes some minor improvements in other parts of the code.

### DIFF
--- a/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
+++ b/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
@@ -131,7 +131,7 @@ public class GenerateTask extends DefaultTask {
     public List<String> jackson2Modules;
     public Logger.Level loggingLevel;
     public boolean disableTrailingSemiColon;
-    public Map<Class<?>, List<String>> injectCustomProperties;
+    public Map<String, List<String>> injectCustomProperties;
 
     private Settings createSettings(URLClassLoader classLoader) {
         final Settings settings = new Settings();
--- a/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
+++ b/typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
@@ -31,6 +31,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.Task;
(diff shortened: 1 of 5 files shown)


panchenko:
@manikandan-ravikumar Could you please elaborate on the reasons for `disableTrailingSemiColon`?

manikandan-ravikumar:
@panchenko I am not disabling semi-colons. I am just introducing an option to disable it. The default value for disabling semi-colon is false. 
The reason behind this is, not all typescript projects use the trailing semi-colon pattern as with JavaScript/Typescript semi-colons are optional. Providing an option to disable it will avoid unnecessary manual work.

typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
```diff
@@ -129,6 +130,8 @@ public class GenerateTask extends DefaultTask {
     public boolean jackson2ModuleDiscovery;
     public List<String> jackson2Modules;
     public Logger.Level loggingLevel;
+    public boolean disableTrailingSemiColon;
+    public Map<Class<?>, List<String>> injectCustomProperties;
```
panchenko, line 134: This task is setup in `build.gradle`, but the corresponding user-defined classes are not accessible there.
Should be class name as string - similar to other fields.

manikandan-ravikumar:
> > The reason behind this is, not all typescript projects use the trailing semi-colon pattern as with JavaScript/Typescript semi-colons are optional. Providing an option to disable it will avoid unnecessary manual work.
> 
> I am trying to understand why it matters. Code analysis should be just disabled for the generated file. For example, in the file header there are comments to disable tslint/eslint.

I agree from the functional POV. But from code maintenance POV, it makes this file stand-out. I am working in a project where we maintain a common standard across the code base and it involves no semi-colon usage. So, we had to `generate.ts` and make manual changes on top of it every time. This breaks the concept of having a "generated" file.

typescript-generator-gradle-plugin/src/main/java/cz/habarta/typescript/generator/gradle/GenerateTask.java
manikandan-ravikumar: Addressed this comment

michaelboyles:
"_I am working in a project where we maintain a common standard across the code base and it involves no semi-colon usage_" Then you should simply not try and apply standards to automatically generated code.

Adding options like this just opens a can of worms because there are infinite ways it can violate your standards. Next you will complain we need `maxLineLength`, and so on.

If you really care about generated code matching your standards, you can automatically run linter fixes on it after you generate it. Best solution is just to disable linting for generated code.

manikandan-ravikumar:
I agree with the suggestion and closing this PR. Actually, I have found an alternative way to achieve this :)
Actually, this is not a complain thread. I just opened a PR for a feature I thought was missing. But I agree that functionality in this package can be easily extended using custom type processors and model transformers. 

Hence closing this PR. Thanks for your time and inputs. Appreciate it.

Issue #597, opened 2020-12-11
generateInfoJson and outputKind "global"


achimmihca:
### Scenario

Consider the project structure:

Root
| - Base
| - Leaf

where Leaf references Base. We are using the typescript-generator plugin already in Base.
Leaf holds different data classes. Thereby it also references classes from Base.

Example:
```
// In Base
public enum BaseEnum { YES, NO; }
public class BaseData { public String name; public BaseEnum baseEnum; }

// In Leaf
public class LeafData { public List<BaseData> baseDatas;}
```

So, for BaseData and BaseEnum there are already generated TS declarations.

### Problem

I wanted to do a second run of the typescript-generator plugin in Leaf.
Problem is that the interfaces and enums of Base are declared again. The second run knows nothing about the first run in Base.
TypeScript does not accept duplicate declarations of the enums (the duplicate interfaces would be ok because they are the same).

As workaround, I could name all already generated declarations in "customTypeMappings" configuration of the second run.
However, the first generated file has about 10000 lines and I would need to name some 1000 declarations.
Thus, this not a practical solution for me.

As I understand, [generateInfoJson](http://www.habarta.cz/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html#generateInfoJson) could solve this because it names declarations that already have been generated.
I would want to generate this file in the first run and reference it in the second run (and possibly further runs).

Sadly, we are not using a module system (yet) for out TS code, such that everything is in global scope.
At the moment, generateInfoJson can only be used with outputKind "module" but I need it with "global".

Would this be possible?
Thanks for the help!

vojtechhabarta:
I think that since you have everything in one global scope you could just use typescript-generator output from Leaf module and ignore output from Base (or even not generate it at all). If you have more modules like Leaf1 and Leaf2 you could create new module that would depend on Leaf1 and Leaf2 and that module would generate complete output in one run. Or do you have some reason why to generate the output in multiple runs?

achimmihca:
> Or do you have some reason why to generate the output in multiple runs?

The example is simplified.
In general (and in my case), the classes in Base and Leaf are different. Leaf uses (i.e. builds upon) a part of Base, but Base also provides different classes.
The goal is to generate a TS interface for all of them.

Our project structure is not optimal (it has grown historically).
Nonetheless, I think this feature would a good addition to typescript-generator.

achimmihca:
[...]

PR #1143, opened 2026-07-16
Ensure that the Maven plugin configuration keeps working

### LINKED ISSUE #1140: Cannot create instance of class cz.habarta.typescript.generator.ModuleDependency in version 4.1.1
I started using version 4.1.1 and when compiling get the following error:

`Unable to parse configuration of mojo cz.habarta.typescript-generator:typescript-generator-maven-plugin:4.1.1:generate for parameter dependency: Cannot create instance of class cz.habarta.typescript.generator.ModuleDependency: cz.habarta.typescript.generator.ModuleDependency.<init>()`

with version 4.0.0 it compiles without problems

timothyjward:
The Maven Plugin was using the core ModuleDependency type as part of its configuration, which requires that the type has a zero-argument constructor. When this constructor was removed it broke the Maven Plugin configuration. This commit adds a type directly to the Maven Plugin, ensuring that the future evolution of ModuleDependency does not break the Maven Plugin and vice-versa

Fixes #1140


typescript-generator-maven-plugin/src/main/java/cz/habarta/typescript/generator/maven/ModuleDependency.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.File;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * A simple data class representing a module dependency for the TypeScript
+ * generator Maven plugin. It decouples the ModuleDependency class from the core 
+ * TypeScript generator library, allowing foreasier maintenance and potential future
+ * changes in the core library without being limited by the Maven plugin's requirements.
+ */
+public class ModuleDependency {
```
kkuegler, line 14: It seems that both the `peerDependency` (issue #829, initially called `usePeerDeps`) and `global` (issue #597) options from the core `ModuleDependency` class are actually configuration that can be specified via Maven, but the documentation does not cover them yet.

If we want to decouple Maven and core via separate classes, these should be included here, too.

typescript-generator-maven-plugin/src/main/java/cz/habarta/typescript/generator/maven/GenerateMojo.java
kkuegler, line 1: PR #1150 tackles a similar topic, but for Gradle. It touches `ModuleDependency`, and also a couple more config classes. Did you happen to check if any of the other classes is problematic in Maven, too?

timothyjward: I didn’t check any other types - this fix was enough to get the maven plugin working for me, and not being hugely familiar with the project I didn’t attempt a broader fix. I’m happy for this PR to be used as the basis of a fuller fix if that is what is needed, or for it to be merged as the first step in that direction.

typescript-generator-maven-plugin/src/main/java/cz/habarta/typescript/generator/maven/ModuleDependency.java
timothyjward: As undocumented properties I wasn’t aware of them. Having a separate configuration type for the maven plugin going forward may actually be helpful here. It would clarify which items are actually supported configuration properties for the plugin, distinct from the library.

